PR #16, opened 2022-01-03
Generate available Platform.sh commands at build time

### LINKED ISSUE #13: Only install platform.sh tools when it's actually needed
The cli will download the platform.sh PHAR when it's started for the first time:
```
Download additional CLI
```

I don't need platform.sh and probably never will. It would be great if this download only triggers the first time you actually need to run some cloud commands.

That way, people that only use the cli for local development, never have to download it (and waste energy doing that).

Now I get the reason why the PHAR is downloaded on first time. It's because the Platform.sh CLI is started to retrieve the commands it has, and then those commands are appended as cloud:* commands to the symfony cli. But that could also be solved by doing something like `Use 'symfony help cloud' to view all cloud commands` that would then trigger the download to retrieve the additional commands.

I could also live with a pragmatic solution: add an option to `~/.symfony5/config.json` that can be used to disable the cloud completely. That way, we can automate our development environment to always set it.

fabpot:
Closed #13

This PR achieves a lot but it does 2 main things:

 * It moves the parsing of the Platform.sh commands at build time (that avoids any overhead at runtime)
 * It now downloads the Platform.sh binary only the first time a `cloud:*` command is run (so, never downloaded if you are not using Platform.sh)


local/platformsh/generator/commands.go
```diff
@@ -0,0 +1,192 @@
...
+		if len(aliases) > 0 {
+			aliasesAsString += "\n\t\tAliases:  []*console.Alias{\n"
+			for _, alias := range aliases {
+				aliasesAsString += "\t\t\t" + alias + ",\n"
+			}
+			aliasesAsString += "\t\t},"
+		}
+		hide := ""
+		if command.Hidden {
+			hide = "		Hidden:   console.Hide,"
```
stof, line 181: shouldn't this start with a newline char ?

fabpot: Indeed, it makes the code more readable, fixed in 85d9470

Issue #10, opened 2022-01-01
Terminal output misaligned


94noni:
There is a misalign on the table output for `symfony server:list`

Please see screenshot
<img width="1313" alt="Capture d’écran 2022-01-01 à 13 35 19" src="https://user-images.githubusercontent.com/1358361/147850688-aa096fa5-76b8-4146-91ac-bbe3772bce44.png">

I've not tested all commands,
Also I am not fluent with Go, so for now I am "unable" to try to fix it :)
Thanks

---

On macOS 12.1 (21C52)

```bash
symfony version
Symfony CLI version 5.0.4 (c) 2017-2022 Symfony SAS (2022-01-01T10:25:37Z - stable)
```

Surfoo:
symfony-cli generate a hyperlink for the port, but the package tablewriter doesn't handle them in automatic padding and count the whole string, not only the text `8000`. 
Here a test : https://go.dev/play/p/P6M-6O8dtqu and the same code on my terminal :  
![test](https://user-images.githubusercontent.com/553659/147862735-a1f61d86-e4f5-4c1e-8483-cc70998950e9.png)

A quick fix should be to remove the link in symfony-cli. A better solution might be to fix the problem in the tablewriter package (not me)

94noni:
@Surfoo ho I didn't saw this on mine (macOS Terminal.app)
Indeed it could have gave a clue before :), I ll see if I can play with this binary locally and improve it

Surfoo:
You could remove the link on the port and add a column with only the hyperlink without the href formatter.

PR #19, opened 2022-01-03
improvement management of the links

Surfoo:
see #10 

It looks like this now: 
screen

### DIFF
--- a/commands/local_server_list.go
+++ b/commands/local_server_list.go
@@ -20,7 +20,10 @@
 package commands
 
 import (
+	"fmt"
 	"sort"
+	"strconv"
+	"strings"
 
 	"github.com/olekukonko/tablewriter"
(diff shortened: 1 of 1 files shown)


tucksaun:
thank you @Surfoo for taking the time for submitting a PR.

we specifically implemented support for ANSI terminal links in the formatting code in order to show links here so removing them would feel like moving backward 😞 
but I agree this is not ideal so I tried to fix it in the upstream library: https://github.com/olekukonko/tablewriter/pull/206
let's see how it goes there 🤞

tucksaun:
And unfortunately, I'm not sure the upstream PR will get merged soon as the last merged PR dates from March 2021 and the last release from February 2021 but I might have something to workaround this in the meantime: #224

fabpot:
Closing in favor of #224

PR #22, opened 2022-01-04
Add .php-version for new projects created with --php

1ed:
I'm new to Go and this is my naive try to implement this https://github.com/symfony/cli/issues/491 proposal

> I think when creating a new project with the --php option like --php 8.1 a .php-version file containing the given version should be added to the project, so when I run symfony php it will use the correct version not the default.

### DIFF
--- a/commands/local_new.go
+++ b/commands/local_new.go
@@ -146,6 +146,12 @@ var localNewCmd = &console.Command{
 			return err
 		}
 
+		if "" != c.String("php") && !c.Bool("cloud") {
+			if err := createPhpVersionFile(c.String("php"), dir); err != nil {
+				return err
+			}
+		}
+
 		if !c.Bool("no-git") {
 			if _, err := exec.LookPath("git"); err == nil {
 				if err := initProjectGit(c, s, dir); err != nil {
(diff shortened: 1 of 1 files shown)


commands/local_new.go
```diff
@@ -371,3 +377,19 @@ func forcePHPVersion(v, dir string) (string, error) {
...
 }
+
+func createPhpVersionFile(v, dir string) error {
+	file := filepath.Join(dir, ".php-version")
+	f, err := os.OpenFile(file, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	if err != nil {
+		return errors.Wrapf(err, "unable to create %s", file)
+	}
+	if _, err = f.WriteString(v + "\n"); err != nil {
+		f.Close()
```
fabpot, line 388: If there was an error, the file has not been open, so it cannot be closed, right?

fabpot:
I'm not sure we should do it by default.
I would certainly not do it when using `--cloud` for instance as the PHP version is already part of the `.platform.app.yaml` file that we are using to select the right PHP version.
Also, if we generate this file, should we commit it to Git automatically?

1ed:
Thanks for your review!

> I would certainly not do it when using --cloud for instance as the PHP version is already part of the .platform.app.yaml file that we are using to select the right PHP version.

I've modified the code.

> Also, if we generate this file, should we commit it to Git automatically?

I've put it right before the initial git commit, so this file will be committed, and I think it should be, as if someone has multiple PHP versions on their machine this is better for DX.

IMHO the best solution would be to set the correct PHP version only 1 place, in composer.json, but currently it's not supported by symfony-cli and platform.sh neither, AFAIK or maybe we should detect conflicting versions.

commands/local_new.go
1ed: That error was handled before right after `os.OpenFile`, at this point the file should be open, this is about handling write errors or I'm missing something? Maybe it's too verbose and I should use `defer`, but I've read not to (https://www.joeshaw.org/dont-defer-close-on-writable-files/). Maybe in this case it would be good enough though.

fabpot:
Thank you

PR #48, opened 2022-01-18
Remove some Platform.sh aliases to avoid confusion

fabpot:


ruudk:
Why is Platform.sh even integrated? I don't get it, it's so specific. Just one of many options.

fabpot:
Because we have a tight integration with them for which we get a small portion of the revenue to finance Symfony. That's the "new version" of SymfonyCloud.

fabpot:
Think of Platform.sh as being the "official PaaS for Symfony"

PR #61, opened 2022-01-23
added webPath flag to open:local

### LINKED ISSUE #44: Add an option to `open:local` command to open website with custom path
Hello ! 👋🏻

I'm actuality using API Platform and I've created a Makefile for automate some process.
When i run `symfony open:local` I would like to add an option to specify the default path I want to open. 
Ex : `/api/docs` by default in API Platform. Maybe the command can be `symfony open:local --webPath=/api/docs` ?

I never coded in Go, that's why I ask for this functionality. 😅 

Thanks a lot ! 😄

henrikac:
This PR attempts to add the `--webPath` flag that was requested in #44.

Running `symfony open:local --webPath=/api/docs` will open the browser on `{scheme}://127.0.0.1:{port}/api/docs`

Should close #44.

### DIFF
--- a/commands/openers.go
+++ b/commands/openers.go
@@ -62,7 +62,7 @@ var projectLocalOpenCmd = &console.Command{
 		url := fmt.Sprintf("%s://127.0.0.1:%d", pidFile.Scheme, pidFile.Port)
 		webPath := c.String("webPath")
 		if webPath != "" {
-			url = fmt.Sprintf("%s/%s", url, webPath)
+			url = fmt.Sprintf("%s%s", url, webPath)
 		}
 		abstractOpenCmd(url)
 		return nil
--- a/commands/openers.go
+++ b/commands/openers.go
@@ -21,6 +21,7 @@ package commands
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/skratchdot/open-golang/open"
 	"github.com/symfony-cli/console"
@@ -46,7 +47,7 @@ var projectLocalOpenCmd = &console.Command{
 	Flags: []console.Flag{
 		dirFlag,
 		&console.StringFlag{
-			Name:  "webPath",
+			Name:  "path",
 			Usage: "Default path which the project should open on",
 		},
 	},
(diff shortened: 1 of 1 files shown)


commands/openers.go
```diff
@@ -45,6 +45,10 @@ var projectLocalOpenCmd = &console.Command{
 	Usage:    "Open the local project in a browser",
 	Flags: []console.Flag{
 		dirFlag,
+		&console.StringFlag{
+			Name:  "webPath",
```
rimi-itk, line 49: This flag should probably be called `web-path` to match the convention of other flags, e.g. `--document-root`.

commands/openers.go
```diff
@@ -55,7 +59,12 @@ var projectLocalOpenCmd = &console.Command{
 		if !pidFile.IsRunning() {
 			return console.Exit("Local web server is down.", 1)
 		}
-		abstractOpenCmd(fmt.Sprintf("%s://127.0.0.1:%d", pidFile.Scheme, pidFile.Port))
+		url := fmt.Sprintf("%s://127.0.0.1:%d", pidFile.Scheme, pidFile.Port)
+		webPath := c.String("webPath")
+		if webPath != "" {
```
rimi-itk, line 64: A minor thing: The recommended way to write this in Go is

```go
if webPath := c.String("webPath"); webPath != "" { … }
```

cf. https://go.dev/doc/effective_go#if.

commands/openers.go
```diff
@@ -55,7 +59,12 @@ var projectLocalOpenCmd = &console.Command{
...
+			url = fmt.Sprintf("%s%s", url, webPath)
```
rimi-itk, line 65: We should probably make sure that a slash is added after the port number, i.e. write something like

```go
fmt.Sprintf("%s/%s", url, strings.TrimLeft(webPath, "/"))
```
(cf. https://pkg.go.dev/strings#TrimLeft)

rimi-itk (commented):
Looks good to me. Some suggestions added.

Arkanii:
Oh ! Thanks a lot @henrikac ! 😍

commands/openers.go
tucksaun: `c.String("webPath")` should always return a string (worst case an empty one)
so you can probably simplify even more and remove the if, like so:
`abstractOpenCmd(fmt.Sprintf("%s://127.0.0.1:%d/%s", pidFile.Scheme, pidFile.Port, strings.TrimLeft(c.String("webPath"))))`

rimi-itk: This will leave us with a trailing slash if the `webPath` flag is not set, but that's not an issue.

commands/openers.go
fabpot: Agreed

fabpot: maybe `--path` is enough here.

henrikac:
Thank you all for the feedback! I hope this new commit will fix all of the minor issues mentioned :)

rimi-itk (approved):

tucksaun (approved):

PR #65, opened 2022-01-24
removed dependency on go-homedir

henrikac:
This PR removes the direct dependency on [go-homedir](https://github.com/mitchellh/go-homedir/tree/v1.1.0) and uses Go's built-in [os.UserHomeDir()](https://pkg.go.dev/os#UserHomeDir) function instead.

The tests still passes `go test ./...` as before but with one less dependency :)

### DIFF
--- a/commands/platformsh.go
+++ b/commands/platformsh.go
@@ -25,7 +25,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/mitchellh/go-homedir"
 	"github.com/symfony-cli/console"
 	"github.com/symfony-cli/symfony-cli/local/php"
 	"github.com/symfony-cli/symfony-cli/local/platformsh"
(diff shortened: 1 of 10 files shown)


fabpot:
Not sure about this one.

First, in the README:

"Why not just use os/user? The built-in os/user package requires cgo on Darwin systems. This means that any Go code that uses that package cannot cross-compile. But 99% of the time the use for os/user is just to retrieve the home directory, which we can do for the current user without cgo. This library does that, enabling cross-compilation."

As we are not using cgo, that's important to us.

But more importantly, this code has been battled-tested with Vagrant and many different edge cases. It's a very small dep anyway, so I think we should keep it.

henrikac:
I don't see this in the README - is this in a different README?

This PR is using the `UserHomeDir` function from the `os` package and not `os/user` but I don't mind closing the PR again if you prefer to keep `go-homedir` :)

Kocal:
In this README: https://github.com/mitchellh/go-homedir/tree/v1.1.0#go-homedir

henrikac:
I see. This was written before the `os.UserHomeDir` function was added to Go.

PR #85, opened 2022-02-04
Disable Go module proxy

hslatman:


.goreleaser.yml
```diff
@@ -34,8 +34,8 @@ builds:
 
 # proxies from the go mod proxy before building
 # https://goreleaser.com/customization/gomod
-gomod:
-  proxy: true
+#gomod:
+#  proxy: true
```
fabpot, line 38: Can you remove the 4 lines instead?

hslatman: Done

PR #90, opened 2022-02-09
Don't ignore --p12 in server:start

pschultz:
server:start ignored the --p12 flag. With this PR server:start uses the specified file instead of generating a certificate when --p12 is given.


pschultz:
The failing check seems unrelated to this change.

PR #137, opened 2022-05-13
Rework PATH_INFO computation

### LINKED ISSUE #135: PATH_INFO is wrong when URL is having query parameters
Hi, 

Start a local server (I'm using php-fpm), and request the following URL: `https://127.0.0.1:8000/index.php/foo?foo=bar`

You will get `PATH_INFO => /foo?foo=bar`. This bug also corrupts the `PHP_SELF` env variable (will be `/index.php/foo?foo=bar`, which is the first bug that led me to the `PATH_INFO` error) because PHP is using `PATH_INFO` to determine `PHP_SELF`:

https://github.com/php/php-src/blob/9a90bd705483004c2ef408ee9e9bb0902beade3f/sapi/fpm/fpm/fpm_main.c#L559-L577

I think the bug resides here: 

https://github.com/symfony-cli/symfony-cli/blob/672a739d3c307518dc2c06754db3829f6b5e241c/local/php/envs.go#L40-L46

`pathInfo` seems to take everything in `req.RequestURI` after filename **including query params**.

tucksaun:
Before we were adding PATH_INFO only when `.php` was detected in the URI.
Now we always provide the PATH_INFO based on what the Go web server parsed
(which excludes the query string) and if we detect `.php` we only provide
the trailing bit after.

This will remove any query string from PATH_INFO (fix #135), and will also
always provide a PATH_INFO even if the script name is not within the URI.

### DIFF
--- a/local/php/envs.go
+++ b/local/php/envs.go
@@ -36,12 +36,12 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
 		https = "On"
 	}
 
-	pathInfo := ""
-	if pos := strings.Index(strings.ToLower(req.RequestURI), ".php"); pos != -1 {
-		file := req.RequestURI[:pos+4]
+	pathInfo := req.URL.Path
+	if pos := strings.Index(strings.ToLower(pathInfo), ".php"); pos != -1 {
+		file := pathInfo[:pos+4]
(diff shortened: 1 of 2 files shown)


nlemoine:
Thanks!

PR #138, opened 2022-05-13
Update supported Platform.sh services

tucksaun:
@fabpot would you be keen to accept a PR automating this process, with a github action ran once in a while for example?


fabpot:
@tucksaun Sure, that would be great.

PR #141, opened 2022-05-16
Add check for docker compose plugin

### LINKED ISSUE #140: Docker Compose detection
Docker Compose has been moved to a Docker CLI plugin, so the check for a `docker-compose` executable for the book requirements: https://github.com/symfony-cli/symfony-cli/blob/48ca923fa3adc2e725788626f02501dff52ddc6a/book/reqs.go#L140 is outdated.

ArnaudLier:
Fixes #140

Maybe you want to check for it differently?

### DIFF
--- a/book/reqs.go
+++ b/book/reqs.go
@@ -137,7 +137,9 @@ func CheckRequirements() (bool, error) {
 	}
 
 	// Docker Compose
-	if _, err := exec.LookPath("docker-compose"); err != nil {
+	_, composeBinary := exec.LookPath("docker-compose")
+	composePlugin := exec.Command("docker", "compose").Run()
+	if composeBinary != nil && composePlugin != nil {
 		ready = false
 		terminal.Println("<error>[KO]</> Cannot find Docker Compose, please install it <href=https://docs.docker.com/compose/install/>https://docs.docker.com/compose/install/</>")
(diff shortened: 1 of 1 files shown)


book/reqs.go
```diff
@@ -137,7 +137,9 @@ func CheckRequirements() (bool, error) {
 	}
 
 	// Docker Compose
-	if _, err := exec.LookPath("docker-compose"); err != nil {
+	_, composeBinary := exec.LookPath("docker-compose")
+	composePlugin := exec.Command("docker", "compose").Run()
```
fabpot, line 141: Can't we check the version of the `docker` binary instead?

ArnaudLier: Presence of the docker binary does not mean you have the `compose` subcommand.

ArnaudLier: Oh yeah, but you can have the latest Docker CLI and no compose plugin?

tucksaun: Indeed. And docker compose is still available as a standalone binary as well.

fabpot: oh, right, this is a plugin.

tucksaun:
The usages of `docker-compose` in `book/checkout.go` will probably needs some updates as well

PR #142, opened 2022-05-16
Add automatic update of the generated code

tucksaun:
Following the discussion in #138

With this PR the generated code is checked for update every four hours from 8 to 18 UTC every weekday.
If a change is detected, tests are run and if they pass the updated code is committed and pushed.
(please note this push will **not** trigger the CI).
The workflow can also be triggered manually (`workflow_dispatch`) if needed.

I also disabled the fact that the workflow fails for pull requests when the code is not updated because there's a high chance that the code gets updated between the PR opening and the fact that it got merged.
Nonetheless, the check is kept for the releasing workflow and the code is also updated within each PR (but not committed) to ensure it is up-to-date before running the tests.

Finally, some cache on dependencies has been added to speed up the check that will be run several times a day.

@fabpot let me know if this is OK for you or if you need some tweaks.

### DIFF
--- a/.github/workflows/go_generate_update.yml
+++ b/.github/workflows/go_generate_update.yml
@@ -1,0 +1,50 @@
+name: Update generated code
+
+on:
+    workflow_dispatch: ~
+    schedule:
+        - cron:  '0 8-18/4 * * 1-5'
+
+jobs:
+    update:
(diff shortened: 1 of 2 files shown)


fabpot:
Fantastic!
Thank you.

PR #155, opened 2022-07-07
Append a trailing slash to Elasticsearch URL

### LINKED ISSUE #154: Generated ELASTICSEARCH_URL seems to lack necessary trailing slash
I'm using what seems to be the default config from the FriendsOfSymfony [Elastica bundle recipe](https://github.com/symfony/recipes-contrib/blob/main/friendsofsymfony/elastica-bundle/5.0/config/packages/fos_elastica.yaml):

```
# .../config/packages/fos_elastica.yaml
fos_elastica:
    clients:
        default: { url: '%env(ELASTICSEARCH_URL)%' }
```

I'd expected this to work with the Docker automatic config, but it seems that although the public port (8856) is picked up correctly from my docker config, the ELASTICSEARCH_URL that's generated is `ELASTICSEARCH_URL=http://127.0.0.1:8856` (I see this with `symfony var:export --multiline`.)

Unfortunately, it seems that fos_elastica expects the URL to have a trailing slash. With the above config, every request gives me a "Malformed URL" HttpException. If I manually change the url to exactly `http://127.0.0.1:8856/`, with the trailing slash, everything works fine.

I've worked around it by switching the config to use the host and port variables that symfony-cli also successfully sets up:

```
fos_elastica:
    clients:
        default:
            host: '%env(ELASTICSEARCH_HOST)%'
            port: '%env(ELASTICSEARCH_PORT)%'
```

...and adding defaults for them to my .env, so I still end up pointed to http://localhost:9200/ by default in other, non-Docker/symfony-cli environments. (It also works if I change my config to `default: { url: '%env(ELASTICSEARCH_URL)%/' }`, of course.)

There seem to be a few moving parts here, and I'm not sure I'm not missing something, but it feels to me like the generated URL should have the trailing slash if the config generated by adding fos_elastica requires it? (Or is ELASTICSEARCH_URL not specifically targeted at that bundle?) Any insight appreciated!

tucksaun:
Fix #154

@fabpot instead of just always appending the trailing slash I went with something different because Platform.sh relationship definition includes a nil path component (see https://docs.platform.sh/add-services/elasticsearch.html#relationship) and I don't want the CLI to break production of your users that could have implemented a workaround for this bug in production.

But if you prefer we can just always append a trailing slash.

### DIFF
--- a/envs/docker.go
+++ b/envs/docker.go
@@ -375,6 +375,7 @@ func (l *Local) dockerServiceToRelationship(client *docker.Client, container typ
 				"host":   host,
 				"ip":     host,
 				"port":   formatDockerPort(p.PublicPort),
+				"path":   "/",
 				"rel":    "elasticsearch",
 				"scheme": "http",
 			}
--- a/envs/envs.go
+++ b/envs/envs.go
@@ -211,7 +211,11 @@ func extractRelationshipsEnvs(env Environment) Envs {
 				values[fmt.Sprintf("%sNAME", prefix)] = endpoint["path"].(string)
 				values[fmt.Sprintf("%sDATABASE", prefix)] = endpoint["path"].(string)
 			} else if rel == "elasticsearch" {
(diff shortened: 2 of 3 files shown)


fabpot:
Thank you

PR #167, opened 2022-08-04
Restore bash installer and adapt it for Symfony CLI v5

tucksaun:
Over the past months, I have seen several projects using the previous official Symfony CLI installer in CI, pipelines or Dockerfiles.
Since Symfony CLI open sourcing, the installer has not been updated but the repository changed which means those projects are still installing Symfony CLI v4.

This PR brings back the bash installer to this project and adapts it for v5.

Please note we have three small BC breaks here that could potentially impact some pipelines:
- We now require `tar` instead of `gzip`: `tar` being a common tool I think it's reasonable to ask users to install it instead of `gzip`
- The installer now puts the installer into `$HOME/.symfony5`: users can override this behavior using the `--install-dir` option if they need to.
- The installer now installs v5 instead of v4: this could introduce some issues but it's hard to predict exactly how.

### DIFF
--- a/installer/bash-installer
+++ b/installer/bash-installer
@@ -1,0 +1,229 @@
+#!/usr/bin/env bash
+# Symfony CLI installer.
+set -euo pipefail
+
+CLI_LATEST_VERSION_URL="https://get.symfony.com/cli/LATEST"
+CLI_CONFIG_DIR=".symfony"
+CLI_EXECUTABLE="symfony"
+CLI_TMP_NAME="$CLI_EXECUTABLE-"$(date +"%s")
+CLI_NAME="Symfony CLI"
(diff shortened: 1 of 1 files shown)


fabpot:
Great! Thank you @tucksaun

oleg-andreyev:
This introduced issue for all docker image we had, before symfony binary was installed into  $HOME/.symfony and we used to execute symfony server:start as following:

```
      /home/ubuntu/.symfony/bin/symfony server:ca:install &&
      /home/ubuntu/.symfony/bin/symfony server:start
```

after re-building image, everything was broken.

Had to apply: `curl -sS https://get.symfony.com/cli/installer | bash -s -- --install-dir=/home/ubuntu/.symfony/bin`

PR #171, opened 2022-08-15
Symfony Cloud does't support Postgresql 14

omerta:
Hi!

I am studying *The Fast Track* book, I execute:

`symfony new guestbook version=6.0 --php=8.1 --webapp --docker --cloud`

Then, we have the fallowing files:

```
//.env
DATABASE_URL="postgresql://app:!ChangeMe!@127.0.0.1:5432/app?serverVersion=14&charset=utf8"
```

```
//docker-compose.yml
    image: postgres:${POSTGRES_VERSION:-14}-alpine
```

```
//.platform/services.yaml
    type: postgresql:14
```

And finally, I run:

```
$ symfony cloud:deploy
Are you sure you want to push to the main (production) branch? [Y/n] y
Pushing HEAD to the existing environment main

Validating submodules

Validating configuration files

E: Error parsing configuration files:
- services.database.type: "postgresql:14" is not a valid service type.
```
I suggest a solution, I hope it will be useful. Any comment will be welcome.

### DIFF
--- a/commands/local_new_test.go
+++ b/commands/local_new_test.go
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2021-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


commands/local_new_test.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		"testdata/postgresql/10/": {
+			Name:    "database",
+			Type:    "postgresql",
+			Version: "10",
+		},
+		"testdata/postgresql/14/": {
+			Name:    "database",
+			Type:    "postgresql",
+			Version: "13",
```
cafferata, line 41: Your directory string is out of sync. The Docker compose shows also version 14. I think it's more readable/makes more sense when they are in sync. What do you think?

```suggestion
			Version: "14",
```

fabpot:
Great! Thank you so much.

PR #189, opened 2022-10-05
Enable the local proxy as a service on macOS

tucksaun:
I always forget to start it after restart so I propose we make the Formula create the service automatically

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -95,6 +95,9 @@ brews:
       type: optional
     install: |-
       bin.install "symfony"
+    service: |-
+      run ["#{bin}/symfony", "local:proxy:start", "--foreground"]
+      keep_alive true
 
 nfpms:
   - file_name_template: '{{ .ConventionalFileName }}'


fabpot:
Makes sense. That will work for Linux as well I suppose.

tucksaun:
I suppose so if one is using homebrew.
But I never used homebrew on Linux so I can't be sure.

fabpot:
Thank you.

tucksaun:
Regarding Linux, and for the record, support would need more work:

Basically one would need to:
1. add required files (mainly service definition but also script to enable the service) in this repo
2. configure GoReleaser to include files in the packages and run scripts after install (see https://goreleaser.com/customization/nfpm/)

From my experience (implementing this back then for Blackfire's agent packaging) step 1 requires quite some work and some time to be implemented properly and cross platform.
But if someone wants to work on this it looks like nFPM team did a good job on writing a good starting point: https://nfpm.goreleaser.com/tips/?h=service

Issue #194, opened 2022-10-09
Installer generate wrong download url


jeje-ch:
The installer generates `https://github.com/symfony-cli/symfony-cli/releases/latest/download/symfony-cli_linux_arm64.tar.gz` which return a 404; using `https://github.com/symfony-cli/symfony-cli/releases/download/v5.4.15/symfony-cli_5.4.15_linux_arm64.tar.gz` works fine.

tucksaun:
It's not that the installer generated a wrong URL, it's more the other way around: the archives named changed in 64487345cb4808fb525d151d7cd2ec51d1112a14.

@fabpot is it possible to revert 64487345cb4808fb525d151d7cd2ec51d1112a14?
currently the installer has not way of determining the version to build the filename :/

jakubtobiasz:
Hi 🙋🏼‍♂️!

Totally agree. PHP Setup action by @shivammathur is dead, also `https://get.symfony.com/cli/installer` fails in my docker images 💀.

mtovmassian:
Experiencing the same issue in our CI pipeline. Is there a quick temporary workaround to deliver anyway?

jakubtobiasz:
@mtovmassian 

If you use Setup PHP action:
Look at our temporary solution in Sylius' builds -> https://github.com/Sylius/Sylius/pull/14425

If you use Docker:
Replace the installer with the solution from above.

oallain:
Hello,

IMO, this issue is fixed with [v5.4.16](https://github.com/symfony-cli/symfony-cli/releases/tag/v5.4.16).

Regards.

jakubtobiasz:
Yes, it works. Thanks, @fabpot for the fast fix 🎉!

fabpot:
Fixed now with 5.4.16.

PR #196, opened 2022-10-10
[BugFix] Changed archive name

jakubtobiasz:
Hi 🙋🏼‍♂️!

// fixes: #194 

Previously, every release included packages for every architecture without a specified version, so going to `https://github.com/symfony-cli/symfony-cli/releases/latest/download/symfony-cli_linux_arm64.tar.gz` invariably resulted in downloading the latest version. https://github.com/symfony-cli/symfony-cli/commit/64487345cb4808fb525d151d7cd2ec51d1112a14 changed it and now every asset includes the version in its name so it is impossible to predict the latest version to download.

This PR might be considered as a temporary (but simultaneously very important at this moment) fix and with the knowledge we have we might find a solution to how to handle it.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -33,7 +33,7 @@ builds:
       - -trimpath
 
 archives:
-  -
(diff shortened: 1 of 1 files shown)


fabpot:
Fixed in upcoming version.

Issue #170, opened 2022-08-11
Gzip support


aszenz:
Some our app routes send big json responses (>50MB), which fail on browsers without gzipping, with gzipping it's only like (5MB), problem is locally with symfony server we are unable to test those routes.
It would be great to have gzip support, optionally ofcourse

aszenz:
https://github.com/nytimes/gziphandler seems like a nice way to do it

PR #198, opened 2022-10-11
feat: add gzip for local webserver

shyim:
I develop currently very often in Gitpod. If an application has huge JS it takes very long to load. Setting up caddy is annoying.

So I just added simple gzip support. It's currently enabled by default when the client supports it using that nytimes library

Also relates to #170

### DIFF
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -67,7 +67,7 @@ var localServerStartCmd = &console.Command{
 		&console.BoolFlag{Name: "no-humanize", Usage: "Do not format JSON logs"},
 		&console.StringFlag{Name: "p12", Usage: "Name of the file containing the TLS certificate to use in p12 format"},
 		&console.BoolFlag{Name: "no-tls", Usage: "Use HTTP instead of HTTPS"},
-		&console.BoolFlag{Name: "use-gzip", Usage: "Use GZIP", DefaultValue: false},
+		&console.BoolFlag{Name: "use-gzip", Usage: "Use GZIP"},
 	},
 	Action: func(c *console.Context) error {
 		ui := terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin)
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -67,6 +67,7 @@ var localServerStartCmd = &console.Command{
 		&console.BoolFlag{Name: "no-humanize", Usage: "Do not format JSON logs"},
 		&console.StringFlag{Name: "p12", Usage: "Name of the file containing the TLS certificate to use in p12 format"},
 		&console.BoolFlag{Name: "no-tls", Usage: "Use HTTP instead of HTTPS"},
+		&console.BoolFlag{Name: "use-gzip", Usage: "Use GZIP", DefaultValue: false},
 	},
 	Action: func(c *console.Context) error {
 		ui := terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin)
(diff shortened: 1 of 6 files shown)


tucksaun:
IIRC gzipping dynamic content is not recommended because of BREACH/CRIME so I don't know if this is a good idea to enable this by default.

Also, I'm curious about actual performance: Gziping is usually useful when the combo deflating, transferring, and inflating content is faster than transferring raw content. This is easy to achieve on slow networks but is it on localhost?
Do you have any numbers so that we can check the impact?

shyim:
Yea it makes I guess more sense to have it opt in. I will look like how to get that option from the cli to the actual server :sweat_smile: 

It makes a really huge difference when you run it on an external server and load the  page. In my case the unmodified build js file is alomost 60mb. With gzip enabled it's "just" 10mb.

tucksaun:
Oh, you are using a remote server! it makes sense then 👍 

But yeah I would make this option opt-in as I'm not sure this is optimal for the main use case that is using the web server locally.
I can help you with this if you wish.

shyim:
How would you like to get that cli option there. I am not sure familiar with this code base

tucksaun:
You would need to first add the flag to the command in `commands/local_server_start.go`, then use the flag to propagate it to the `Config` (defined in `local/project/config.go`), and then finally to the HTTP Server (defined in `local/http/http.go` and configured in `local/project/project.go`)

shyim:
I did it :)

commands/local_server_start.go
```diff
@@ -67,6 +67,7 @@ var localServerStartCmd = &console.Command{
 		&console.BoolFlag{Name: "no-humanize", Usage: "Do not format JSON logs"},
 		&console.StringFlag{Name: "p12", Usage: "Name of the file containing the TLS certificate to use in p12 format"},
 		&console.BoolFlag{Name: "no-tls", Usage: "Use HTTP instead of HTTPS"},
+		&console.BoolFlag{Name: "use-gzip", Usage: "Use GZIP", DefaultValue: false},
```
tucksaun, line 70: Default `DefaultValue` for `BoolFlag` is false so you can omit it

local/http/http.go
```diff
@@ -66,8 +79,22 @@ func (s *Server) Start(errChan chan error) (int, error) {
 	}
 	s.serverPort = strconv.Itoa(port)
 
+	gzipWrapper, err := gziphandler.GzipHandlerWithOpts(gziphandler.ContentTypes(gzipContentTypes))
```
tucksaun, line 82: `gzipWrapper` should be initialized only when `s.UseGzip` is true

local/http/http.go
```diff
@@ -66,8 +79,22 @@ func (s *Server) Start(errChan chan error) (int, error) {
...
+
+	if err != nil {
+		return port, err
```
tucksaun, line 85: You need to wrap `err` with `errors.WithStack(err)` so that we can get stack traces in debug mode

local/http/http.go
```diff
@@ -51,13 +52,25 @@ type Server struct {
...
+	"text/html",
+	"text/plain",
+	"text/csv",
+	"text/javascript",
+	"text/css",
+	"application/json",
+	"application/javascript",
+	"application/vnd.api+json",
+}
+
```
tucksaun, line 73: If we want to restrict the gzipped content type I think we should at least add the following:
- text/xml 
- image/svg+xml

and even maybe those ones:
- application/atom+xml
- application/rss+xml

shyim:
I fixed all the comments :)

PR #200, opened 2022-10-14
Fix `TestParseDockerComposeServices` test

tucksaun:
This test was too sensible to PSH services version upgrades so the CI is currently broken.
Also, this means services versions are not updated by the CI.

@fabpot but in order to make it possible to allow this I had to enable environment variables expansion in Docker compose files during `local:new` so I did another commit so that it makes it to the changelog.


fabpot:
Thank you.

PR #218, opened 2022-11-13
feat: fpm sapi listen always on sockets

### LINKED ISSUE #213: Start PHP-FPM as socket
Hey,

I would like to have a mode / always serve the PHP-FPM as socket. On GitHub Codespaces / Gitpod / Jetbrains Gateway a listing port will be shown in a modal to forward that port and this is mostly not wanted for the php-fpm specific port.

So I would like to get rid of listing to ports, instead use a socket file. It also feels clearer to not use a random port 🤔

shyim:
Fixes #213 

Cloud based dev environments show all local listing ports and wants to forward them. But for PHP-FPM it's useless 😅 

PHP-FPM is only available on Unix, so I didn't build a flag or so.

## Before

image

## After

image

### DIFF
--- a/local/php/php_server.go
+++ b/local/php/php_server.go
@@ -32,6 +32,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"runtime"
 	"strconv"
 	"strings"
@@ -46,6 +47,7 @@ import (
 	"github.com/symfony-cli/symfony-cli/local/html"
 	"github.com/symfony-cli/symfony-cli/local/pid"
 	"github.com/symfony-cli/symfony-cli/local/process"
+	"github.com/symfony-cli/symfony-cli/util"
 )
 
 // Server represents a PHP server process (can be php-fpm, php-cgi, or php-cli)
@@ -98,11 +100,18 @@ func (p *Server) Start(ctx context.Context, pidFile *pid.PidFile) (*pid.PidFile,
 	var binName, workerName string
 	var args []string
 	if p.Version.IsFPMServer() {
+		socketDir := path.Join(util.GetHomeDir(), name(p.projectDir))
+
+		if _, err := os.Stat(socketDir); os.IsNotExist(err) {
+			os.MkdirAll(socketDir, os.ModePerm)
+		}
+
+		p.addr = path.Join(util.GetHomeDir(), name(p.projectDir), "php-fpm.sock")
 		fpmConfigFile := p.fpmConfigFile()
 		if err := ioutil.WriteFile(fpmConfigFile, []byte(p.defaultFPMConf()), 0644); err != nil {
 			return nil, nil, errors.WithStack(err)
 		}
-		pathsToRemove = append(pathsToRemove, fpmConfigFile)
+		pathsToRemove = append(pathsToRemove, fpmConfigFile, p.addr)
 		binName = "php-fpm"
 		workerName = "PHP-FPM"
 		args = []string{p.Version.ServerPath(), "--nodaemonize", "--fpm-config", fpmConfigFile}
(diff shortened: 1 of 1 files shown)


local/php/php_server.go
```diff
@@ -98,11 +100,18 @@ func (p *Server) Start(ctx context.Context, pidFile *pid.PidFile) (*pid.PidFile,
 	var binName, workerName string
 	var args []string
 	if p.Version.IsFPMServer() {
+		socketDir := path.Join(util.GetHomeDir(), name(p.projectDir))
+
+		if _, err := os.Stat(socketDir); os.IsNotExist(err) {
+			os.MkdirAll(socketDir, os.ModePerm)
+		}
+
+		p.addr = path.Join(util.GetHomeDir(), name(p.projectDir), "php-fpm.sock")
```
tucksaun, line 109: I would like to see introduced a `fpmSocketFile` method (similar to `fpmConfigFile`)
And also with the same naming strategy that includes the FPM version

local/php/php_server.go
```diff
@@ -224,8 +233,14 @@ func (p *Server) serveFastCGI(env map[string]string, w http.ResponseWriter, r *h
...
-		if fcgi, err = fcgiclient.Dial("tcp", p.addr); err == nil {
-			break
+		if p.Version.IsFPMServer() {
+			if fcgi, err = fcgiclient.Dial("unix", p.addr); err == nil {
+				break
+			}
+		} else {
+			if fcgi, err = fcgiclient.Dial("tcp", p.addr); err == nil {
+				break
+			}
```
tucksaun, line 243: I'm wondering if we should not refactor this to have a simpler dialer internal function here.
wdyt @fabpot ?

tucksaun (commented):
I like the idea.

Though it probably needs more hardening in some edge cases.
For example, if FPM is still running in the background because of some bugs, the current code is able to run by just picking a new port.

shyim:
@tucksaun do you have an idea how to make the socket bulletproof that it gets stopped 🤔. I didn't encountered a problem yet it stopped smoothly 🤔

tucksaun:
if it always stops smoothly it means we did our job well with @fabpot :D 

More seriously, if you want to simulate this situation: `kill -9` the Symfony CLI binary process and remove the PID file associated with FPM (not the one in $HOME/.symfony5/var but in the subdirectory instead).
You will end up with an FPM process that the CLI is not aware of.

Currently, this means there's a 'ghost' process, but at least restarting the web server will spawn a new FPM process and it will work for the end user.
With a fixed listen address, the user gets this kind of error:
```
$ /Users/tucksaun/Work/src/github.com/symfony-cli/symfony-cli/symfony-cli serve
Following Web Server log file (/Users/tucksaun/.symfony5/log/2b298b2015661d81fcd2d5dcb5ddbe7e15d7b653.log)
Following PHP-FPM log file (/Users/tucksaun/.symfony5/log/2b298b2015661d81fcd2d5dcb5ddbe7e15d7b653/53fb8ec204547646acb3461995e4da5a54cc7575.log)
{"level":"error","time":"2022-11-24T17:26:15-05:00","message":"Unable to start PHP-FPM"}

 [ERROR] PHP-FPM failed to start:
 unable to bind listening socket for address '127.0.0.1:54292': Address already in use (48)
 FPM initialization failed



  PHP server exited unexpectedly: command "PHP-FPM" failed early: exit status 78
```

shyim:
image

and the entire tree is gone for me 🤔 
image

shyim:
So my idea would be:

- We enable pid in php-fpm
- Before php-fpm starting, look for that pid file and get the pid and kill it and cleanup that file

What do you think? I guess it should solve it

tucksaun:
I have the feeling that would complexify the server management code as it would introduce a special case for FPM.

The philosophy so far has been to clean up a maximum so that the CLI does not leak things but if on a rare occasion it does this is not a big deal as it is not written for production workloads.
So with that in mind, I would probably add a random seed as a suffix to the FPM socket path to be sure the process can always start.
As such, in case of failure, the best case is if the PID file and FPM are still there, the current code will kill the server before going on, but if the PID file or the FPM is not there the CLI will still be able to start.

wdyt?

Issue #215, opened 2022-11-11
Feature request: Ability to specify git branch name


andyexeter:
It'd be great if there was a way to specify the branch name created when created a project. Currently it is hardcoded to `main`:

https://github.com/symfony-cli/symfony-cli/blob/6d0e50b31e618b48e696dc243f4ac0179b3d2da5/git/init.go#L28

I think there should be a command line option to change the default, e.g:

```
symfony new --webapp --git-branch="v1"
```

PR #220, opened 2022-11-14
Add git branch option

andyexeter:
Implementation of my feature request in #215

This PR adds a `--git-branch` option to the `new` command, allowing developers to override the default `main` branch name.


wouterj:
What about, instead of adding a flag, using `init.defaultBranch` ?

tucksaun:
I guess it has been done to simplify SymfonyCloud/Platform.sh onboarding as well as simplifying starting working with the Fast Track.
Do you confirm @fabpot ?

If that is the case maybe we can force the branch in those cases and rely on the default git config otherwise ?

fabpot:
@tucksaun You're right.

tucksaun:
@fabpot 👍 

what about this alternative approach then: #235 ?

PR #224, opened 2022-11-16
Fix table formatting when ANSI terminal links are present in the table

### LINKED ISSUE #10: Terminal output misaligned

tucksaun:
Alternative to #19 
Fix #10

### DIFF
--- a/commands/tablewriter_patch_ansi.go
+++ b/commands/tablewriter_patch_ansi.go
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2021-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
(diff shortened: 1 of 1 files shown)


tucksaun:
wdyt @fabpot ?
too hacky for you?

fabpot:
Works for me as the comment is very clear about why we're doing it for now. Thank you.

PR #231, opened 2022-11-24
Expose the CLI version and binary name ...

tucksaun:
… when running `console`, `php` and `composer` wrappers

Exposing that info via environment variables will allow a better Symfony Maker experience, see https://github.com/symfony/maker-bundle/pull/1238

```
$ ./symfony-cli php -r "echo getenv('SYMFONY_CLI_BINARY_NAME').' '.getenv('SYMFONY_CLI_VERSION');"
symfony-cli dev
$
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,6 +43,13 @@ var (
 	buildDate string
 )
 
+func getCliExtraEnv() []string {
+	return []string{
+		"SYMFONY_CLI_VERSION=" + version,
+		"SYMFONY_CLI_BINARY_NAME=" + console.CurrentBinaryName(),
+	}
+}
+
 func main() {
 	args := os.Args
 	name := console.CurrentBinaryName()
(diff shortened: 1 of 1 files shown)


tucksaun:
waiting for @weaverryan and @jrushlow feedback before moving to ready

OskarStark:
I would be more expressive:
```diff
- SYMFONY_CLI_BIN_NAME
+ SYMFONY_CLI_BINARY_NAME
```

or only `SYMFONY_CLI_BINARY` ?

Do you have an example of the value?

tucksaun:
The usual value would be `symfony` but if someone renames the binary to `symfony-cli` then it would be `symfony-cli`

fabpot (approved):

fabpot:
Is it still draft? Looks great to me.

PR #233, opened 2022-11-25
feat: add `local:proxy:url` command

Kocal:
Hi,

I need to get the proxy's URL in order to export `HTTP_PROXY`/`HTTPS_PROXY` env vars, for running Blackfire, Blackfire Player, and Cypress on our local Symfony app.

ATM, I either hardcode it, either I fetch from the file `~/.symfony5/proxy.json` with the following command, which is not ideal:  
```shell
$(cat ~/.symfony5/proxy.json | jq -r '(.host + ":" + (.port|tostring))')
```

I've implemented a new command `local:proxy:get-url` which:
- `exit 1` if the proxy is not started
image

- returns the URL only, if the proxy is started 
image

WDYT? 
Thanks.

### DIFF
--- a/commands/local_proxy_url.go
+++ b/commands/local_proxy_url.go
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2021-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package commands
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/symfony-cli/console"
+	"github.com/symfony-cli/symfony-cli/local/pid"
+	"github.com/symfony-cli/terminal"
+)
+
+var localProxyUrlCmd = &console.Command{
+	Category: "local",
+	Name:     "proxy:url",
+	Aliases:  []*console.Alias{{Name: "proxy:url"}},
+	Usage:    "Get the local proxy server URL",
+	Description: `Get the local proxy server URL, for example if you need to define HTTP_PROXY/HTTPS_PROXY environment variables
+when running an external program:
+
+e.g. with Blackfire: 
+	HTTP_PROXY=$(symfony proxy:url) HTTPS_PROXY=$(symfony proxy:url) blackfire curl ...
+	
+e.g. with Cypress: 
+	HTTP_PROXY=$(symfony proxy:url) HTTPS_PROXY=$(symfony proxy:url) ./node_modules/bin/cypress ...
+`,
+	Action: func(c *console.Context) error {
+		pidFile := pid.New("__proxy__", nil)
+		if !pidFile.IsRunning() {
+			return errors.New("The proxy server is not running")
+		}
+
+		url := fmt.Sprintf("%s://127.0.0.1:%d", pidFile.Scheme, pidFile.Port)
+		terminal.Print(url)
+
+		return nil
+	},
+}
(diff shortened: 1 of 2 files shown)


tucksaun:
Looks interesting to be able to get get the proxy URL programmatically 👍 
Though I think I would prefer a dedicated command.

Kocal:
Yeah a dedicated command would be more clear and easier to maintain, I'm gonna change it asap.

commands/local_proxy_get_url.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/symfony-cli/console"
+	"github.com/symfony-cli/symfony-cli/local/pid"
+	"github.com/symfony-cli/terminal"
+)
+
+var localProxyGetUrlCmd = &console.Command{
+	Category: "local",
+	Name:     "proxy:get-url",
+	Aliases:  []*console.Alias{{Name: "proxy:get-url"}},
```
henrikac, line 34: Suggestion: name it `proxy:url` instead of `proxy:get-url` - just looks cleaner and less typing :)

commands/local_proxy_get_url.go
```diff
@@ -0,0 +1,51 @@
...
+	Usage:    "Get the local proxy server URL",
+	Flags: []console.Flag{
+		dirFlag,
+	},
+	Action: func(c *console.Context) error {
+		pidFile := pid.New("__proxy__", nil)
+		if !pidFile.IsRunning() {
+			terminal.Println("    <error>Not Running</>")
+			os.Exit(1)
```
tucksaun, line 43: you can directly return an error here, `symfony-cli/console` will take care of displaying the error with a nice red error box (à la Symfony) on stderr and have an exit code of 1

commands/local_proxy_get_url.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/symfony-cli/terminal"
+)
+
+var localProxyGetUrlCmd = &console.Command{
+	Category: "local",
+	Name:     "proxy:get-url",
+	Aliases:  []*console.Alias{{Name: "proxy:get-url"}},
+	Usage:    "Get the local proxy server URL",
+	Flags: []console.Flag{
+		dirFlag,
```
tucksaun, line 37: unused flag?

commands/local_proxy_get_url.go
tucksaun: `proxy:url` sound better indeed.

commands/local_proxy_get_url.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/symfony-cli/console"
+	"github.com/symfony-cli/symfony-cli/local/pid"
+	"github.com/symfony-cli/terminal"
+)
+
+var localProxyGetUrlCmd = &console.Command{
+	Category: "local",
+	Name:     "proxy:get-url",
+	Aliases:  []*console.Alias{{Name: "proxy:get-url"}},
+	Usage:    "Get the local proxy server URL",
```
tucksaun, line 35: Can you add some `Description` to explain the use case? thanks

commands/local_proxy_get_url.go
Kocal: Hum yes

Kocal:
Comments have been addressed.

commands/local_proxy_url.go
```diff
@@ -0,0 +1,56 @@
...
+e.g. with Blackfire: 
+	HTTP_PROXY=$(symfony proxy:url) HTTPS_PROXY=$(symfony proxy:url) blackfire curl ...
+	
+e.g. with Cypress: 
+	HTTP_PROXY=$(symfony proxy:url) HTTPS_PROXY=$(symfony proxy:url) ./node_modules/bin/cypress ...
+`,
+	Action: func(c *console.Context) error {
+		pidFile := pid.New("__proxy__", nil)
+		if !pidFile.IsRunning() {
+			return errors.New("Not Running")
```
tucksaun, line 48: what about "The proxy server is not running" ?

tucksaun (commented):
except a small remark it looks good. thanks!

commands/local_proxy_url.go
Kocal: Ok to update it in the `proxy:status` command too?

tucksaun: the `proxy:status` already shows a "Local Proxy Server" (kind-of) banner and is more meant to be run interactively so I think it's fine for now as you are less likely to find this message in script logs.

Kocal: LGTM

tucksaun (approved):

fabpot:
@Kocal Thank you

Can you add a note in the docs about this new feature? You can reuse the 2 examples as they seem like the best usage examples of this new command.

Kocal:
Opened at https://github.com/symfony/symfony-docs/pull/17479

PR #235, opened 2022-11-25
fix: only force Git main branch name for new project when --cloud is used

tucksaun:
overseeds #220

### DIFF
--- a/commands/local_new.go
+++ b/commands/local_new.go
@@ -320,7 +320,7 @@ func initProjectGit(c *console.Context, s *terminal.Spinner, dir string) error {
 	terminal.Println("* Setting up the project under Git version control")
 	terminal.Printfln("  (running git init %s)\n", dir)
 	// Only force the branch to be "main" when running a Cloud context to make
-	// om boarding simpler.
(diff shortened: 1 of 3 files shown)


commands/local_new.go
```diff
@@ -319,7 +319,9 @@ func parseDockerComposeServices(dir string) []*CloudService {
 func initProjectGit(c *console.Context, s *terminal.Spinner, dir string) error {
 	terminal.Println("* Setting up the project under Git version control")
 	terminal.Printfln("  (running git init %s)\n", dir)
-	if buf, err := git.Init(dir, c.Bool("debug")); err != nil {
+	// Only force the branch to be "main" when running a Cloud context to make
+	// om boarding simpler.
```
cafferata, line 323:
```suggestion
	// onboarding simpler.
```

cafferata (changes requested):

commands/local_new.go
fabpot, line 323:
```suggestion
	// on-boarding simpler.
```

fabpot (changes requested):

PR #245, opened 2022-12-20
feat(code-quality): add golangci-lint to CI and lint project

darkweak:
Hello, this PR is here to have a better code quality/code style using `golangci-lint`. It adds the CI job to ensure every other PR will be as well written as possible.

### DIFF
--- a/.github/workflows/go_generate_update.yml
+++ b/.github/workflows/go_generate_update.yml
@@ -35,6 +35,9 @@ jobs:
                 run: |
                     RESULT=$(git status --untracked-files=no --porcelain)
                     echo "gitstatus=$RESULT" >> $GITHUB_OUTPUT
+            - 
+                name: golangci-lint
+                uses: golangci/golangci-lint-action@v3
             -
                 name: Test
                 if: steps.git.outputs.gitstatus != ''
--- a/.github/workflows/go_generate_update.yml
+++ b/.github/workflows/go_generate_update.yml
@@ -35,9 +35,6 @@ jobs:
                 run: |
                     RESULT=$(git status --untracked-files=no --porcelain)
                     echo "gitstatus=$RESULT" >> $GITHUB_OUTPUT
-            - 
-                name: golangci-lint
-                uses: golangci/golangci-lint-action@v3
             -
                 name: Test
                 if: steps.git.outputs.gitstatus != ''
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,11 @@
+run:
+  timeout: 30s
+  issues-exit-code: 1
+
+linters:
+  disable:
+    # TODO remove these linters and fix the deadcode/errcheck/unused/varcheck
+    - deadcode
+    - errcheck
+    - unused
+    - varcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,9 +3,14 @@ run:
   issues-exit-code: 1
 
 linters:
-  disable:
-    # TODO remove these linters and fix the deadcode/errcheck/unused/varcheck
-    - deadcode
-    - errcheck
-    - unused
-    - varcheck
+  enable:
+    - gci
+    - wrapcheck
+
+linters-settings:
+  wrapcheck:
+    ignorePackageGlobs:
+      # We already make sure your own packages wrap errors properly
+      - github.com/symfony-cli/*
+  errcheck:
+      ignore: github.com/symfony-cli/terminal:^(Ep|P)rint
--- a/book/checkout.go
+++ b/book/checkout.go
@@ -35,7 +35,7 @@ import (
 func (b *Book) Checkout(step string) error {
 	// FIXME: keep vendor/ node_modules/ around before git clean, but them back as they will be updated the right way, less Internet traffic
 	// FIXME: if the checkout is to a later step, no need to remove the DB, we can just migrate it
-	os.Chdir(b.Dir)
+	_ = os.Chdir(b.Dir)
 	step = strings.Replace(step, ".", "-", -1)
 	tag := fmt.Sprintf("step-%s", step)
 	branch := "work-" + tag
@@ -117,7 +117,7 @@ func (b *Book) Checkout(step string) error {
 	}
 
 	printBanner("<comment>[WEB]</> Stopping the Local Web Server", b.Debug)
-	executeCommand([]string{"symfony", "server:stop"}, b.Debug, true, nil)
+	_ = executeCommand([]string{"symfony", "server:stop"}, b.Debug, true, nil)
 
 	printBanner("<comment>[WEB]</> Stopping the Platform.sh tunnel", b.Debug)
 	if err := executeCommand([]string{"symfony", "tunnel:close", "-y"}, b.Debug, true, nil); err != nil {
@@ -131,7 +131,7 @@ func (b *Book) Checkout(step string) error {
 
 	printBanner("<comment>[SPA]</> Stopping the Local Web Server", b.Debug)
 	if _, err := os.Stat(filepath.Join(b.Dir, "spa")); err == nil {
-		executeCommand([]string{"symfony", "server:stop", "--dir", filepath.Join(b.Dir, "spa")}, b.Debug, true, nil)
+		_ = executeCommand([]string{"symfony", "server:stop", "--dir", filepath.Join(b.Dir, "spa")}, b.Debug, true, nil)
 	} else {
 		terminal.Println("Skipped for this step")
 	}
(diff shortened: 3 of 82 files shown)


.github/workflows/go_generate_update.yml
```diff
@@ -35,6 +35,9 @@ jobs:
                 run: |
                     RESULT=$(git status --untracked-files=no --porcelain)
                     echo "gitstatus=$RESULT" >> $GITHUB_OUTPUT
+            - 
+                name: golangci-lint
+                uses: golangci/golangci-lint-action@v3
```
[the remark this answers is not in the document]
tucksaun: we could enable `only-new-issues` in order to not overwhelm new contributors with code not up to date with quality

tucksaun (changes requested):
Thank you @darkweak for working on this 🙏
This has been on my mind for a while 👍

Just a couple of thoughts:

- I believe we should tweak the settings to match our use case though.
For example, `unusedresult` should probably be disabled (at least for several functions from `terminal`).

- We should also take the opportunity to some of those issues instead of masking them (not checking errors for example) which makes me believe it would be better to add the configuration in this PR, maybe fix the obvious issues (`ioutil`-> `os` for example) and fix the other issues with other PRs.
 
any thought @fabpot ?

.github/workflows/go_generate_update.yml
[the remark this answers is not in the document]
darkweak: You're right ✌️

darkweak:
> I believe we should tweak the settings to match our use case though.
For example, unusedresult should probably be disabled (at least for several functions from terminal).

IMHO if that's unused that should be removed 😄 

> We should also take the opportunity to some of those issues instead of masking them

That's the first iteration and was to implement asap the golangci-lint. I plan to add some error handling (for those I just disabled checks).

.github/workflows/go_generate_update.yml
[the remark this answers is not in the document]
darkweak: After thinking about that, 
> we could enable only-new-issues in order to not overwhelm new contributors with code not up to date with quality

If they rebase their PR they'll have the correct code.

tucksaun:
> IMHO if that's unused that should be removed 😄

I'm not referring to unused but unused results that are not always useful and I personally find having tons of `_, _  = ` unreadable (see all the changes for functions as `terminal.Printfln`)

> That's the first iteration and was to implement asap the golangci-lint. I plan to add some error handling (for those I just disabled checks).

That is the reason why I mentioned enabling only-new-issues. By doing so we can more quickly merge this PR that would only add configuration and most obvious fix and then we can fix other issues in dedicated PRs

darkweak:
Oh sorry, I misunderstood what you said then, thank you for your explanations.
I'll try to find time to update the PR during this week-end 👍

darkweak:
@tucksaun the PR is ready.

darkweak:
@fabpot @tucksaun any update?

.golangci.yml
```diff
@@ -0,0 +1,11 @@
...
+  timeout: 30s
+  issues-exit-code: 1
+
+linters:
+  disable:
+    # TODO remove these linters and fix the deadcode/errcheck/unused/varcheck
+    - deadcode
+    - errcheck
+    - unused
+    - varcheck
```
tucksaun, line 11: we can remove `deadcode`, `varcheck` and `unused` from the list

tucksaun: also missing new line at the end of the file

book/checkout.go
```diff
@@ -35,7 +35,9 @@ import (
 func (b *Book) Checkout(step string) error {
 	// FIXME: keep vendor/ node_modules/ around before git clean, but them back as they will be updated the right way, less Internet traffic
 	// FIXME: if the checkout is to a later step, no need to remove the DB, we can just migrate it
-	os.Chdir(b.Dir)
+	if err := os.Chdir(b.Dir); err != nil {
+		return err
```
tucksaun, line 39: We usually wrap errors (with `errors.WithStack`) to get nice stack traces

tucksaun: this also means `wrapcheck` should probably be enabled in golangci configuration

.golangci.yml
tucksaun: and regarding errcheck, given the fact that you fixed most of then we can enable errcheck and disable it only for terminal functions like `Println`

tucksaun: (see my other configuration suggestion)

tucksaun, line 11:
```suggestion
linters:
  enable:
    - gci
    - wrapcheck

linters-settings:
  wrapcheck:
    ignorePackageGlobs:
      # We already make sure your own packages wrap errors properly
      - github.com/symfony-cli/*
  errcheck:
      ignore: github.com/symfony-cli/terminal:^(Ep|P)rint
```

tucksaun (changes requested):

darkweak:
Ping @tucksaun

.golangci.yml
```diff
@@ -0,0 +1,16 @@
...
+    - gci
+    - wrapcheck
+
+linters-settings:
+  wrapcheck:
+    ignorePackageGlobs:
+      # We already make sure your own packages wrap errors properly
+      - github.com/symfony-cli/*
+  errcheck:
+      ignore: github.com/symfony-cli/terminal:^(Ep|P)rint
```
tucksaun, line 16: missing newline at the end of the file

book/checkout.go
```diff
@@ -35,7 +35,9 @@ import (
 func (b *Book) Checkout(step string) error {
 	// FIXME: keep vendor/ node_modules/ around before git clean, but them back as they will be updated the right way, less Internet traffic
 	// FIXME: if the checkout is to a later step, no need to remove the DB, we can just migrate it
-	os.Chdir(b.Dir)
+	if err := os.Chdir(b.Dir); err != nil {
+		return errors.WithStack(err)
+	}
```
tucksaun, line 40:
```suggestion
	if err := errors.WithStack(os.Chdir(b.Dir)); err != nil {
		return err
	}
```

I'm not saying we have to do it, but we can do it when there's only a single return value, this also better preserve stack traces IIRC

book/checkout.go
```diff
@@ -110,40 +112,44 @@ func (b *Book) Checkout(step string) error {
...
 		if err := executeCommand(append(dockerComposeBin(), "down", "--remove-orphans"), b.Debug, false, nil); err != nil {
-			return err
+			return errors.WithStack(err)
 		}
 	} else {
 		terminal.Println("Skipped for this step")
 	}
 
 	printBanner("<comment>[WEB]</> Stopping the Local Web Server", b.Debug)
-	executeCommand([]string{"symfony", "server:stop"}, b.Debug, true, nil)
```
tucksaun, line 120: I would tend to think that @fabpot deliberately skip this one

book/checkout.go
```diff
@@ -110,40 +112,44 @@ func (b *Book) Checkout(step string) error {
...
 
 	printBanner("<comment>[GIT]</> Checking out the step", b.Debug)
 	if err := executeCommand([]string{"git", "checkout", "-B", branch, tag}, b.Debug, false, nil); err != nil {
-		return err
+		return errors.WithStack(err)
 	}
 
 	printBanner("<comment>[SPA]</> Stopping the Local Web Server", b.Debug)
 	if _, err := os.Stat(filepath.Join(b.Dir, "spa")); err == nil {
-		executeCommand([]string{"symfony", "server:stop", "--dir", filepath.Join(b.Dir, "spa")}, b.Debug, true, nil)
```
tucksaun, line 134: same

commands/local_proxy_domain_attach.go
```diff
@@ -40,16 +41,16 @@ var localProxyAttachDomainCmd = &console.Command{
 	Action: func(c *console.Context) error {
 		projectDir, err := getProjectDir(c.String("dir"))
 		if err != nil {
-			return err
+			return errors.WithStack(err)
```
tucksaun, line 44: this one should be wrapped in the parent function

commands/local_proxy_domain_attach.go
```diff
@@ -40,16 +41,16 @@ var localProxyAttachDomainCmd = &console.Command{
...
 		}
 
 		homeDir := util.GetHomeDir()
 		config, err := proxy.Load(homeDir)
 		if err != nil {
-			return err
+			return errors.WithStack(err)
```
tucksaun, line 50: same

commands/local_proxy_start.go
```diff
@@ -68,7 +68,7 @@ var localProxyStartCmd = &console.Command{
 			}
 			if err := reexec.Background(varDir); err != nil {
 				if _, isExitCoder := err.(console.ExitCoder); isExitCoder {
-					return err
+					return errors.WithStack(err)
```
tucksaun, line 71: IIRC `console.ExitCoder` errors must not be wrapped as we have special handling for them in `symfony-cli/console`

tucksaun (changes requested):
I'm stopping here because reviewing this kind of PR is complicated 😖.
Also, the PR needs rebasing.
Would you reconsider opening a new PR with only the configuration and then creating one PR per topic fixed (one by one)?
That would help a lot IMO as they are a couple of fixes that could be merged quickly then.

darkweak:
@tucksaun I'll open some PRs to review easier. I'll start with one that add only golangci-lint in the project with the given configuration without noises. The PR should be ready this evening. ✌️

PR #253, opened 2023-01-09
Fix Symfony CLI to work with Platform.sh CLI v4.x

### LINKED ISSUE #252: Warning containing Platform.sh migration message
Running any Symfony Cloud command displays a warning coming from the (legacy) Platform.sh CLI:

```
Warning: Both the Legacy (PHP) CLI and the New CLI are installed on your system.
To migrate fully, delete the Legacy CLI.

Remove the following file completely: /home/xyz/.platformsh/bin/platform
This operation is safe and doesn't delete any data.

Do you want to remove this file now? [Y/n]
```

This message should not be displayed as we're not in the new (Go) CLI context.
This probably comes from the fact that the [Symfony CLI exports a different name for the executable](https://github.com/symfony-cli/symfony-cli/blob/4d2117a776465a7f548db1996f8a104804410c65/commands/platformsh.go#L126-L128).

akalipetis:
* Use PLATFORMSH_CLI_NO_INTERACTION=1 to make sure the migration message is not displayed during installation
* Use PLATFORMSH_CLI_WRAPPED=1 to make sure migration messages are not shown when running the CLI

Fix #252

### DIFF
--- a/commands/platformsh.go
+++ b/commands/platformsh.go
@@ -126,6 +126,7 @@ func (p *platformshCLI) executor(args []string) *php.Executor {
 		"PLATFORMSH_CLI_APPLICATION_NAME=Platform.sh CLI for Symfony",
 		"PLATFORMSH_CLI_APPLICATION_EXECUTABLE=symfony",
 		"XDEBUG_MODE=off",
+		"PLATFORMSH_CLI_WRAPPED=1",
 	}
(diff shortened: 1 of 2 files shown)


fabpot:
I think there is more. In #254, I've switched from the installer to using Github directly to download binaries.

Issue #237, opened 2022-12-06
Index.php in subdirectories [ feature request ]


maciekpaprocki:
First of all, big fan of this tool. It cuts a lot of loading time on mac vs docker and it's very easy to set up. I am pretty much recommending it to everyone who listens ( and a lot of people who don't :) ). 

And here's a problem. Your project is too good to support only modern PHP frameworks. It would be great for it to be able to support WP too as it really requires such a tool. Fortunately, it's really not far from accomplishing that. 

Currently, WP pretty much works out of the box for the front end. It struggles a bit on the backend though. The issue is that all the requests will always go to the index.php in the web unless file.php is appended. 

Example web folder structure ( using bedrock, composer version of WP ):
```yaml
web
  - index.php
  - wp
    - wp-admin
      - index.php
```

When going to / I see web/index.php
When going to /wp/wp-admin/index.php I see web/wp/wp-admin/index.php
When going to /wp/wp-admin/ I weirdly see web/index.php again. 

Allowing the directories paths to go to the root index in them would solve this issue. I think it might be breaking change, but it would be weird that someone would have other files and didn't want to serve them, so maybe it's not. 

This change would probably make behaviour more similar to how apache and Nginx work by default, so possible it would make other old-school frameworks work out of the box. 

I am happy to help with some coding but probably will have to be pointed in the right direction as Go is not my strong suit.

tucksaun:
Let me rephrase your need to be clear: when the requested URL is a directory, the Symfony CLI should check for an `index.php` file, and if one is found use it as the passthru controller, otherwise fallback to the default passthru controller?

If so, I see the use case and it might be easy enough to make it work.
But we might also encounter more edge cases as we need to support FPM but also PHP built-in server.
And what if one does not want this behavior because they are migrating some legacy code exposing PHP scripts to Symfony?
Apache and Nginx have a configurability we don't have to allow users to override this behavior so we have to think about where to draw the line.

@fabpot any thoughts about this?

maciekpaprocki:
Yes, that's exactly it. 

I kind of assumed that everyone is using the server only locally. If not then yes, that might have security implications. That could be specified as a parameter then or YAML config and maybe even disabled by default. 

If it's used locally only then I don't really see much of a problem with even legacy code. They probably want it to run either way. Also, that's probably a more common configuration. From my experience either it's htaccess / nginx working exactly like that or Nginx just hardcoding everything to one index.php. 

[...]

PR #259, opened 2023-02-01
Added support for index.php in subdirectories

maciekpaprocki:
https://github.com/symfony-cli/symfony-cli/issues/237

- Added support for index.php in subdirectories ( without index.php being in the URL ) 
- Added a bunch of tests

It's my first go PR and I wasn't able to find any contributor rules or anything like that, so it's straightforward PR. I know I am going to have to change a few things, so don't worry about bashing it :)


tucksaun (changes requested):
This code might be necessary (but I'm not sure) but is not enough to fix the linked issue because the HTTP handler (https://github.com/symfony-cli/symfony-cli/blob/75f91e31404e/local/http/http.go#L201) will not forward the call to PHP at all.
You first need to make sure the handler detects such PHP scripts and then forward the request to PHP

maciekpaprocki:
Hey. Just to make sure we are talking about this same. This script is now returning the correct values: 

```sh
mkdir -p web/subdirectory
echo "<?php echo 'root'; " > web/index.php
echo "<?php echo 'subdirectory'; " > web/subdirectory/index.php
go run main.go server:start -d
curl localhost:8000 # returns root
curl localhost:8000/subdirectory # returns subdirectory
```

I think the issue you might be talking about though is that command will refuse to enable PHP if the root PHP file is not available. Is that it?

maciekpaprocki:
Actually, i think we are talking about different things. 

Currently the behaviour of a server is that it will ignore PHP all together if there's no root index.php, so i didn't change anything there. 

Going back to your issue. I am crap at GO so forgive me if I say something stupid, but seems that Handler is used for static files and then as a fallback it calls s.callback. The original part of the routing for other .php files is included in envs file, so i guess this one should be too.

local/php/envs.go
```diff
@@ -37,11 +37,30 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
...
 		file := pathInfo[:pos+4]
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
 			scriptName = file
 			pathInfo = pathInfo[pos+4:]
+			filenameMatch = true
+		}
+
+	}
+
+	if pathInfo != "/" && filenameMatch == false {
```
henrikac, line 52: Would this work on Windows as Windows separate paths with `\\`?

maciekpaprocki: Ugh, actually, that line is probably not a problem, it's just small optimisation, but the other ones might be an issue. Frankly I would have to set up windows for it and it's not quick thing to do. Will have a think, maybe some friend can test it for me.

henrikac: or maybe use [filepath.Separator](https://pkg.go.dev/path/filepath#pkg-constants) instead of hardcoding `"/"`? :)

maciekpaprocki: I added a more proper way of dealing with it. Should be windows proof, but then again, i don't have windows :(

local/php/envs.go
```diff
@@ -37,11 +37,30 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
...
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
 			scriptName = file
 			pathInfo = pathInfo[pos+4:]
+			filenameMatch = true
+		}
+
+	}
+
+	if pathInfo != string(os.PathSeparator) && filenameMatch == false {
+		paths := strings.Split(strings.TrimRight(pathInfo[1:], "/"), "/")
```
henrikac, line 53: would this be a candidate for `os.PathSeparator` too? :)

maciekpaprocki: Goddamit. Yes :)

tucksaun:
> Hey. Thanks for quick review. Just to make sure we are talking about this same. This script is now returning the correct values:
> 
> ```shell
> mkdir -p web/subdirectory
> echo "<?php echo 'root'; " > web/index.php
> echo "<?php echo 'subdirectory'; " > web/subdirectory/index.php
> go run main.go server:start -d
> curl localhost:8000 # returns root
> curl localhost:8000/subdirectory # returns subdirectory
> ```
> 
> I think the issue you might be talking about though is that command will refuse to enable PHP if the root PHP file is not available. Is that it?

Hum, sorry I was wrong: I misread the Handler part. Everything is fine there.

maciekpaprocki:
Great. Thanks for reviewing. I am happy for that to go. It would be good to know it works on windows too, but it shouldn't have any adverse effects if it doesn't, the feature just wouldn't be available.

local/php/envs.go
```diff
@@ -37,11 +37,30 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
...
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
 			scriptName = file
 			pathInfo = pathInfo[pos+4:]
+			filenameMatch = true
+		}
+
+	}
+
+	if pathInfo != string(os.PathSeparator) && filenameMatch == false {
+		paths := strings.Split(strings.TrimRight(pathInfo[1:], string(os.PathSeparator)), string(os.PathSeparator))
```
tucksaun, line 53: we are not dealing with filepaths but with URLs here, so we need to split by `/` instead of `os.PathSeparator`

local/php/envs.go
```diff
@@ -37,11 +37,30 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
...
+		pathsLength := len(paths)
+		for i := 0; i < pathsLength; i++ {
```
tucksaun, line 55: in a more go idiomatic way:
```suggestion
		for i, n := 0, len(paths); i < n; i++ {
```

tucksaun, line 55: the loop should probably be reversed so that an `index.php` higher in the tree does not take precedence (this case should probably be added to the tests)

local/php/envs.go
```diff
@@ -37,11 +37,30 @@ func (p *Server) generateEnv(req *http.Request) map[string]string {
...
+			poppedPath := paths[:pathsLength-i]
+			indexDir := string(os.PathSeparator) + filepath.Join(poppedPath...)
+			file := filepath.Join(indexDir, "index.php")
```
tucksaun, line 58: what about using `p.passthru` here?
does it make sense?

local/php/envs.go
tucksaun: `Trim` instead of `TrimRight` is also probably necessary

tucksaun (changes requested):

local/php/envs.go
maciekpaprocki: I don't think i understand it.

local/php/envs.go
maciekpaprocki: It is currently as I just use -i in removal from array, but good catch about the test. Need to add one more nested level to actually be sure.

maciekpaprocki: Also, I am gonna rewrite the loop using pop push as it's actually confusing as hell.

local/php/envs.go
maciekpaprocki: ah ok. It might sense, might be sometimes weird with some complicated passthroughs, but it makes sense.

local/php/envs.go
maciekpaprocki: Actually, golang pop syntax is actually more complicated than current one so no rewrite, but I am adding tests.

local/php/envs.go
maciekpaprocki: Actually it's weird, but changing to trim might actually cause some damage by stripping more slashes from the front which can be problematic. 

Basically, it should work like that:
```
/subdirectory/ -> subdirectory/index.php
//subdirectory/ -> index.php
```

I guess the more common behaviour of apache or nginx would resolve it to root too as // wouldn't be found as a folder and it's probably better that user would figure out their URLs are wrong locally than when deployed. 

However, should definitely add a test case for that.

maciekpaprocki: Goddamit, sorry. I am thinking out loud apparently. Seems that default behaviour when working with paths like `//subidrectory//index.php/something` is actually a bit inconsistent currently, so I am gonna keep this same behaviour where basically // is ignored ( said that, passed as query string )

maciekpaprocki:
Few fixes 
- removed issues with // in URLs 
- fixed few code style issues mentioned
- added more tests for edge cases

local/php/envs.go
```diff
@@ -31,17 +31,36 @@ import (
...
+			filenameMatch = true
+		}
+
+	}
+
+	if filenameMatch == false && pathInfo != "/" && len(pathInfo) > 0 {
+		paths := strings.Split(strings.Trim(pathInfo, "/"), "/")
+		for i, n := 0, len(paths); i < n; i++ {
+			poppedPath := paths[:len(paths)-i]
+			indexDir := filepath.Join(poppedPath...)
```
tucksaun, line 57: I found the following more readable:
```suggestion
		for n := len(paths); n > 0; n-- {
			indexDir := filepath.Join(paths[:n]...)
```

local/php/envs.go
```diff
@@ -31,17 +31,36 @@ import (
...
+			file := filepath.Join(indexDir, p.passthru)
+			if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
+				scriptName = string(os.PathSeparator) + file
+				pathInfo = pathInfo[strings.Index(pathInfo, indexDir)+len(indexDir):]
+				break
+			}
 		}
 	}
```
tucksaun, line 65: would you mind extracting all this into a private function?
early returns would help to have better readability here

local/php/envs.go
```diff
@@ -31,17 +31,36 @@ import (
...
 		file := pathInfo[:pos+4]
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
 			scriptName = file
 			pathInfo = pathInfo[pos+4:]
+			filenameMatch = true
+		}
+
+	}
+
+	if filenameMatch == false && pathInfo != "/" && len(pathInfo) > 0 {
```
tucksaun, line 53:
```suggestion
	if len(pathInfo) > 1 {
```
seems sufficient here (once extracted to a function and thus not need `filenameMatch`)

local/php/envs.go
```diff
@@ -31,17 +31,36 @@ import (
...
+		paths := strings.Split(strings.Trim(pathInfo, "/"), "/")
+		for i, n := 0, len(paths); i < n; i++ {
+			poppedPath := paths[:len(paths)-i]
+			indexDir := filepath.Join(poppedPath...)
+			file := filepath.Join(indexDir, p.passthru)
+			if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
+				scriptName = string(os.PathSeparator) + file
```
tucksaun, line 60: `scriptName` is not only used for the file path but also to populate some web-related environments.
So if it's not entirely accurate I would rather stick with a `"/"` prefix to remain consistent with every other passthru usage we have and not break anything in the PHP userland side.

The end result goes through `filepath.Join` which will use `filepath.Clean` so this should not cause any issue on Windows anyway for PHP (engine side this time).

tucksaun (changes requested):

local/php/envs.go
maciekpaprocki: Good shout!

local/php/envs.go
```diff
@@ -29,20 +29,37 @@ import (
 	"github.com/symfony-cli/symfony-cli/envs"
 )
 
-func (p *Server) generateEnv(req *http.Request) map[string]string {
-	scriptName := p.passthru
-	https := ""
-	if req.TLS != nil {
-		https = "On"
-	}
+func (p *Server) resolveIndexFile(pathInfo string) (string, string) {
```
tucksaun, line 33: extra line to remove

local/php/envs.go
```diff
@@ -29,20 +29,37 @@ import (
...
+				return "/" + file, pathInfo[strings.Index(pathInfo, indexDir)+len(indexDir):]
+			}
+		}
+	}
+
+	return p.passthru, pathInfo
+}
+
+func (p *Server) generateEnv(req *http.Request) map[string]string {
+
```
tucksaun, line 57: extra line to remove

local/php/envs.go
```diff
@@ -29,20 +29,37 @@ import (
...
 
-	pathInfo := req.URL.Path
 	if pos := strings.Index(strings.ToLower(pathInfo), ".php"); pos != -1 {
 		file := pathInfo[:pos+4]
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
-			scriptName = file
-			pathInfo = pathInfo[pos+4:]
+			return file, pathInfo[pos+4:]
 		}
+
```
tucksaun, line 39: extra line to remove

local/php/envs.go
```diff
@@ -29,20 +29,37 @@ import (
...
+	}
+
+	if len(pathInfo) > 1 {
+		paths := strings.Split(strings.Trim(pathInfo, "/"), "/")
+		for n := len(paths); n > 0; n-- {
+			indexDir := filepath.Join(paths[:n]...)
+			file := filepath.Join(indexDir, p.passthru)
+			if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
+				return "/" + file, pathInfo[strings.Index(pathInfo, indexDir)+len(indexDir):]
```
tucksaun, line 48: as we trim/clean the pathinfo, I see a problem here where the CLI would change the pathinfo and indexDir, resulting in a wrong returned pathInfo as we use indexes and length from different cleaned strings.

For example, `"/subdirectory///subdirectory//foo/"` would cause a weird `"tory//foo/"` pathinfo returned.

tucksaun: Using the `LastIndex` and length of `paths[n-1]` seems logical to me
but then some tests fails

tucksaun: ok found it, actually it's because the loop considers empty elements from `paths`
an additional if allows to fix this.
@maciekpaprocki I have a patch with the suggested changes here and the additional test case: see https://github.com/maciekpaprocki/symfony-cli/pull/1

maciekpaprocki:
Thanks for the fix, I think you removed the empty lines too, so I just squashed it and it's ready to go. 

Thank you

local/php/envs.go
```diff
@@ -29,22 +29,39 @@ import (
...
-	scriptName := p.passthru
-	https := ""
-	if req.TLS != nil {
-		https = "On"
-	}
-
-	pathInfo := req.URL.Path
+func (p *Server) resolveScriptName(pathInfo string) (string, string) {
 	if pos := strings.Index(strings.ToLower(pathInfo), ".php"); pos != -1 {
 		file := pathInfo[:pos+4]
```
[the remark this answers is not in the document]
maciekpaprocki: I don't think pathinfo is ever "", it was at least / for me always. Said that, new function is checking for `.php` and then len so don't think that would be an issue. Not sure what else would you mean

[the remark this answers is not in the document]
henrikac: oh wait... I missed the `pos != -1` part - just pretend that I did not ask this 🙈

local/php/envs.go
```diff
@@ -29,22 +29,39 @@ import (
...
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
-			scriptName = file
-			pathInfo = pathInfo[pos+4:]
+			return file, pathInfo[pos+4:]
+		}
+	}
+
+	if len(pathInfo) > 1 {
```
tucksaun, line 40: I'm not sure this is needed anymore 🤔

local/php/envs.go
```diff
@@ -29,22 +29,39 @@ import (
...
+		paths := strings.Split(strings.Trim(pathInfo, "/"), "/")
+		for n := len(paths); n > 0; n-- {
+			if paths[n-1] == "" {
```
tucksaun, line 43: We use `paths[n-1]` several times here so it's probably a good idea to introduce a variable to get better semantic

tucksaun (commented):
Two last minor comments.
Tomorrow I will play a bit with this to check the performance to be sure this patch does not brings a big hit

maciekpaprocki:
Had a quick look with ab:

```sh
mkdir -p web
echo "<?php echo 'root'; " > web/index.php
go run main.go server:start -d
ab -n 7000 -c 20 http://localhost:8000/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo
```

Both benchmarks had pretty much similar footprint, even that this is pretty hardcore case. 

## Examples:

### Before
Percentage of the requests served within a certain time (ms)
  50%      5
  66%      5
  75%      5
  80%      6
  90%      6
  95%      6
  98%      7
  99%      9
 100%    656 (longest request)

 
 ### After
 Percentage of the requests served within a certain time (ms)
  50%      5
  66%      6
  75%      6
  80%      6
  90%      6
  95%      7
  98%      8
  99%      9
 100%    599 (longest request)
 
I run it a bunch of times, and the results were not really showing any major differences. Over 7k requests in both branches had problems, but I guess it would be probably logging overflowing or some mac process kicking in. 

Obviously, it's not really a reliable test, but better than nothing. 


 
 mac pro 
2.4 GHz 8-Core Intel Core i9
64 GB 2667 MHz DDR4

local/php/envs_test.go
```diff
@@ -33,6 +33,7 @@ type PHPFPMSuite struct{}
 var _ = Suite(&PHPFPMSuite{})
 
 func (s *PHPFPMSuite) TestGenerateEnv(c *C) {
+
```
fabpot, line 36: Should be reverted

maciekpaprocki: Done and rebased

local/php/envs.go
```diff
@@ -29,22 +29,38 @@ import (
...
 	if pos := strings.Index(strings.ToLower(pathInfo), ".php"); pos != -1 {
 		file := pathInfo[:pos+4]
 		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
-			scriptName = file
-			pathInfo = pathInfo[pos+4:]
+			return file, pathInfo[pos+4:]
+		}
+	}
+
+	paths := strings.Split(strings.Trim(pathInfo, "/"), "/")
```
tucksaun, line 40: Using `strings.Split` (or `Trim` /`LeftTrim` as far as I can tell) will actually lead to potential directory traversal:
the heading "/" being removed it becomes possible for one to request an URL that will have ".." components interpreted from the document root and leaving the possibility to one to execute a script outside of the document root.

local/php/envs.go
```diff
@@ -29,22 +29,38 @@ import (
...
+	for n := len(paths); n > 0; n-- {
+		pathPart := paths[n-1]
+		if pathPart == "" {
+			continue
+		}
+
+		file := filepath.Join(append(paths[:n], p.passthru)...)
+		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
+			return "/" + file, pathInfo[strings.LastIndex(pathInfo, pathPart)+len(pathPart):]
```
tucksaun, line 49: using `LastIndex` will cause issues when some components are repeated.
For example, using the added test cases, the `"/subdirectory/subdirectory/foo/subdirectory/bar"` URI will only provide `"/bar"` as the path info

tucksaun (changes requested):

local/php/envs.go
maciekpaprocki: I am gonna test it, not sure how golang behaves with ../. I guessed if it's an issue it would be an issue already before with direct PHP file in url.

local/php/envs.go
maciekpaprocki: Ah crap, it might. Might have to revert it to join of array, unless you think of anything smarter.

tucksaun: I might have something but it changes the code a lot

maciekpaprocki:
One more thing @tucksaun. Seen your commit: 
https://github.com/symfony-cli/symfony-cli/commit/7abd242f5228667b334342e89e95638868150a16

And it solves the execution outside of the index, but it will probably accept things like "/foo/../update.php" which is probably not what we want. My code looks crap though with nested if statements :) 

I added it to tests.

tucksaun:
Hey sorry for the delay.
Regarding DOCUMENT_URI, so far we never got any feedback about routing issues with any major PHP framework so I would rather stick with the current logic. That would be another issue anyway IMO.

Going to review the code now

tucksaun:
code looks quite complicated indeed :/

what I drafted last week was something similar to this:
```
func (p *Server) resolveScriptName(pathInfo string) (string, string) {
	if pos := strings.Index(strings.ToLower(pathInfo), ".php"); pos != -1 {
		file := path.Clean(pathInfo[:pos+4])
		if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
			return file, pathInfo[pos+4:]
		}
	}

	for pos := len(pathInfo); pos > 1; {
		// virtually trim trailing slashes from pathinfo
		for pathInfo[pos-1] == '/' {
			pos--
		}

		if file := filepath.Join(pathInfo[:pos], p.passthru); file == p.passthru {
			break
		} else if _, err := os.Stat(filepath.Join(p.documentRoot, file)); err == nil {
			return file, pathInfo[pos:]
		}

		pos = strings.LastIndexByte(pathInfo[:pos], '/')
	}

	return p.passthru, pathInfo
}
```

The idea is to explore the path bits by bits (parts separated by `/` to be more precise) starting from the right end and going upwards so that as soon as we reach a directory where an index.php file exists we stop here.
As we do the look for the `/` ourselves we can easily "trim' or normalize them on the go.
I can submit a PR to your fork if you wish.

Regarding the ".." we don't really care if they allow relative navigation in the path, this will not end up with a path one could have send directly.
The only important bit is that it should not let one go outside of the document root. here, because we do a `filepath.Join` (calling `Clean`) before doing the `filepath.Join` with document root, the path joined with document root is already safe and will not include relative bits.

The same is probably we true for repeated slashes.
As far as I know, because those are valids when dealing with file paths, webservers will accept them. But probably most web frameworks won't. So if we end up compacting the ones that are part of the script filenames I don't really care much.

maciekpaprocki:
I mean, if we don't care about the path support for ".."  and "//" then we can just do a quick break.
 
```go
if( pathInfo != path.Clean(pathInfo) ){
	return p.passthru, pathInfo
}
```

And that will clean a lot of code. Said that, not sure if we shouldn't support it, so people can test their frameworks with it to actually make sure it's resolved. 

When it comes to the loop, I think having a split might be a bit easier for people to figure it out and when we actually add all the edge cases ( sub/index.php/ vs sub/index.php, // and ../ support ) I think it will end up with the similar amount of code. my code currently also starts from the back and moves to the start, just does that from the array.  

I can probably try to optimise a few things, but it would be a line or two tops if we want to keep all the features and I am not sure if it will actually simplify code.

local/php/envs_test.go
```diff
@@ -182,6 +182,138 @@ func (s *PHPFPMSuite) TestGenerateEnv(c *C) {
...
+		{
+			passthru: "/index.php",
+			uri:      "///subdirectory",
+			expected: map[string]string{
+				"PATH_INFO":       "///subdirectory",
+				"REQUEST_URI":     "///subdirectory",
+				"QUERY_STRING":    "",
+				"SCRIPT_FILENAME": testdataDir + "/public/index.php",
+				"SCRIPT_NAME":     "/index.php",
+			},
```
tucksaun, line 260: because `/subdirectory/index.php` exists I guess one would expect this to match it

tucksaun:
> Said that, not sure if we shouldn't support it, so people can test their frameworks with it to actually make sure it's resolved.

I'm not saying we should clean the pathinfo and not support having `..` or repeated `/`  in it, but the part of the URL leading to the script name could happen to be cleaned and we should not care IMO (and the framework behind the script as well): a typical request should not even arrive with some `/../` because most HTTP clients will compact them before sending the request. Some web servers even just plainly reject requests when `/../` is included in the URI.
This also means we are consistent with what will happen if one adds the script filename.
My last review goes with this logic in mind

> When it comes to the loop, I think having a split might be a bit easier for people to figure it out and when we actually add all the edge cases

That's the reason why the code becomes complicated: there are a lot of conditions to deal with edge cases
But IMHO by not using spit we remove most of the edge cases and needs for string manipulations

local/php/envs_test.go
```diff
@@ -182,6 +182,138 @@ func (s *PHPFPMSuite) TestGenerateEnv(c *C) {
...
+		{
+			passthru: "/index.php",
+			uri:      "/subdirectory///subdirectory//foo/",
+			expected: map[string]string{
+				"PATH_INFO":       "///subdirectory//foo/",
+				"REQUEST_URI":     "/subdirectory///subdirectory//foo/",
+				"QUERY_STRING":    "",
+				"SCRIPT_FILENAME": testdataDir + "/public/subdirectory/index.php",
+				"SCRIPT_NAME":     "/subdirectory/index.php",
+			},
```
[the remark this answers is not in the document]
maciekpaprocki: TBH, I did that to preserve the same functionality as currently you have with default PHP file which was that you would get uri like //something. Said that, the non default file had completely different behaviour either way, so either way it will not be consistent. If you say that's good, I am good with it too.

local/php/envs_test.go
```diff
@@ -182,6 +182,138 @@ func (s *PHPFPMSuite) TestGenerateEnv(c *C) {
...
+		{
+			passthru: "/index.php",
+			uri:      "/../index.php",
+			expected: map[string]string{
+				"PATH_INFO":       "/../index.php",
+				"REQUEST_URI":     "/../index.php",
+				"QUERY_STRING":    "",
+				"SCRIPT_FILENAME": testdataDir + "/public/index.php",
+				"SCRIPT_NAME":     "/index.php",
+			},
```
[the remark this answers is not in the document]
maciekpaprocki: Yeah, i did mostly because i seen that curl doesn't resolve it, so although it's possible in browser it is possible in other environments. Said that, I am gonna drop bigger comment with how nginx deals with that in main topic.

maciekpaprocki:
OK. I had time to set up a basic wp nginx example and actually learned that all the paths will be cleaned up by default, which is completely against what I got from docs :) 

I am not sure how it is with other servers, but I found some mentions of path normalisation in apache too. I guess that would mean, that if we want to make the server similar to that of two main ones we should probably always normalise path. That will remove a lot of if statements, but at this same time, it will make it not back-compatible for those not sanitised URLs currently supported. I am not really sure if that's a problem as I can see a small chance someone on purpose used `../` or `//` in their routing. in fact i am sure symfont does it too. 

I am gonna send an update to the code tomorrow with normalising path for all requests.

PR #264, opened 2023-02-01
Add links to every HTTP services found into the Symfony WDT

### LINKED ISSUE #197: Add Kibana to toolbar if detected by Docker integration?
Hi,

I'm wondering if it is possible to add Kibana in the toolbar if it's detected by Docker integration?

It would be nice as Elasticsearch is already handled by the CLI :)

tucksaun:
Fix #197 by providing an even more generic solution.
based on work made in #263, as such Draft until #263 is merged

Currently, this looks like this:
Screenshot 2023-02-01 at 17 00 43
@fabpot @jmsche @javiereguiluz any help on the look is welcome :)

### DIFF
--- a/local/php/toolbar.go
+++ b/local/php/toolbar.go
@@ -88,10 +88,10 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
 	extraLinks := ``
 	if env, err := envs.NewLocal(p.projectDir, terminal.IsDebug()); err == nil {
 		if url, exists := env.FindServiceUrl("mailer"); exists {
-			webmail = fmt.Sprintf(`<b><a class="sf-cli-webmail" href="%s" target="_blank">Webmail</a></b> <span class="sf-toolbar-status sf-toolbar-status-green">Up</span></a>`, url)
+			webmail = fmt.Sprintf(`<b><a class="sf-cli-webmail" href="%s" rel="noopener" target="_blank">Webmail</a></b> <span class="sf-toolbar-status sf-toolbar-status-green">Up</span></a>`, url)
 		}
 		if url, exists := env.FindServiceUrl("amqp"); exists {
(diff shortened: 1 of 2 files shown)


fabpot:
For the design, maybe we can make the title a link and remove the "Open" link (for the new section, but also for the existing one).

jmsche:
Nice idea from @fabpot :) 

If applied, the link should be added into the `<b>` tag, as putting it around this tag would not render nicely.

tucksaun:
I like the idea @fabpot 
What about the status badge? should we keep them or remove all of them?

tucksaun:
New look:
Screenshot 2023-02-02 at 16 24 03

local/php/toolbar.go
```diff
@@ -81,23 +82,32 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
...
 	blackfire := `<span class="sf-toolbar-status sf-toolbar-status-red">Down</span>`
+	extraLinks := ``
 	if env, err := envs.NewLocal(p.projectDir, terminal.IsDebug()); err == nil {
-		values := envs.AsMap(env)
-		if prefix := env.FindRelationshipPrefix("mailer", "http"); prefix != "" {
-			if url, exists := values[prefix+"URL"]; exists {
-				webmail = fmt.Sprintf(`<span class="sf-toolbar-status sf-toolbar-status-green">Up</span>&nbsp;&nbsp;&nbsp;<a class="sf-cli-webmail" href="%s" target="_blank">Open</a>`, url)
-			}
+		if url, exists := env.FindServiceUrl("mailer"); exists {
+			webmail = fmt.Sprintf(`<b><a class="sf-cli-webmail" href="%s" target="_blank">Webmail</a></b> <span class="sf-toolbar-status sf-toolbar-status-green">Up</span></a>`, url)
```
javiereguiluz, line 91: For security reasons, we could add `rel="noopener"` to all `target="_blank"` links. Some browsers like Chrome do that automatically, but others don't do that yet.

See https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/

fabpot: Works for me.

tucksaun: makes sense 👍 
done

Issue #278, opened 2023-02-22
bug with arm raspberry, please fix it


CodingBDX:
hello,
you have hot fix, because don't fint cli symfony on arm

``` > [ 8/19] RUN curl -sS https://get.symfony.com/cli/installer | bash     &&  mv /root/.symfony5/bin/symfony /usr/local/bin/symfony:                                                                                                             
#0 6.296 Symfony CLI installer                                                                                          
#0 6.301                                                                                                                
#0 6.301 Environment check                                                                                              
#0 6.306   [*] cURL is installed
#0 6.310   [*] Tar is installed
#0 6.313   [*] Git is installed
#0 6.342   [*] Your architecture (arm) is supported
#0 6.345 
#0 6.345 Download                                                                                                       
#0 6.347   Downloading https://github.com/symfony-cli/symfony-cli/releases/latest/download/symfony-cli_linux_arm.tar.gz...
#0 6.414   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#0 6.427                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
#0 7.041 curl: (22) The requested URL returned error: 404 
```

PR #193, opened 2022-10-08
Add more builds

fabpot:
Maybe closes #180

PR #279, opened 2023-02-22
Remove arm from installer as it's not supported

fabpot:


tucksaun:
I guess it is related to #278 but also to #193 ?

During my last tests, the CLI works without problem on my RPI 1, so maybe we can figure out how to release ARM builds properly?
I can work on this but I don't have a ARM7 cpu at hand so some help would be required for final validation.

If we don't want to wait a couple of days to see if I can make it work then 👍 to remove it from the installer.

fabpot:
I think we can wait a few days :)

PR #293, opened 2023-03-15
feat(http): add flag/config to allow CORS requests, close #229

### LINKED ISSUE #229: Feature Request - Web Server CORS?
Hello, currently I'm using `symfony serve` command for local development which starts web server for symfony app binded to 127.0.0.1:8000 (and other random port) but whenever I'm making request from react app which is on 127.0.0.1:3000 I'm getting CORS error.

I know that there are bundles to fix this issue but the reason I don't want to install them is that I don't really need CORS on production and don't really want to add additional logic like `if APP_ENV == dev { add cors headers }` it's just for local development, so I thought maybe the ideal place to handle this issue is actually symfony web server itself since it's used specifically for local development? 

Additional option like `symfony serve --disable-cors` would be perfect solution, or you guys think it's not the right place to handle it?

Kocal:
Hi everyone, this PR is a proposal for #229.

A new flag `--allow-cors` has been added to `server:start` command, and to config `http.allow_cors` in `.symfony.local.yaml`.

When allowing CORS, the follows headers are added to each response:
- `Access-Control-Allow-Origin: *`
- `Access-Control-Allow-Methods: *`
- `Access-Control-Allow-Header: *`

image
image

Is there a way to add automated tests for that?

Thanks!

#SymfonyHackday

### DIFF
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -52,6 +52,7 @@ import (
 
 var localWebServerProdWarningMsg = "The local web server is optimized for local development and MUST never be used in a production setup."
 var localWebServerTlsKeyLogWarningMsg = "Logging TLS master key is enabled. It means TLS connections between the client and this server will be INSECURE. This is NOT recommended unless you are debugging the connections."
+var localWebServerAllowsCORSLogWarningMsg = "Cross-origin resource sharing (CORS) is enabled for all requests.\nYou may want to use https://github.com/nelmio/NelmioCorsBundle to have better control over HTTP headers."
 
 var localServerStartCmd = &console.Command{
 	Category:    "local",
@@ -83,6 +84,7 @@ var localServerStartCmd = &console.Command{
 			EnvVars: []string{"SSLKEYLOGFILE"},
 		},
 		&console.BoolFlag{Name: "no-workers", Usage: "Do not start workers"},
+		&console.BoolFlag{Name: "allow-cors", Usage: "Allow Cross-origin resource sharing (CORS) requests"},
 	},
 	Action: func(c *console.Context) error {
 		ui := terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin)
@@ -186,6 +188,10 @@ var localServerStartCmd = &console.Command{
 
 		if config.TlsKeyLogFile != "" {
 			ui.Warning(localWebServerTlsKeyLogWarningMsg)
+		}
+
+		if config.AllowCORS {
+			ui.Warning(localWebServerAllowsCORSLogWarningMsg)
 		}
 
 		lw, err := pidFile.LogWriter()
--- a/local/http/cors.go
+++ b/local/http/cors.go
@@ -21,14 +21,26 @@ package http
 
 import (
 	"net/http"
+
+	"github.com/rs/zerolog"
 )
 
-func corsWrapper(h http.Handler) http.Handler {
+func corsWrapper(h http.Handler, logger zerolog.Logger) http.Handler {
+	var corsHeaders = []string{"Access-Control-Allow-Origin", "Access-Control-Allow-Methods", "Access-Control-Allow-Headers"}
+
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Access-Control-Allow-Origin", "*")
-		w.Header().Set("Access-Control-Allow-Methods", "*")
-		w.Header().Set("Access-Control-Allow-Headers", "*")
+		for _, corsHeader := range corsHeaders {
+			w.Header().Set(corsHeader, "*")
+		}
 
 		h.ServeHTTP(w, r)
+
+		for _, corsHeader := range corsHeaders {
+			if headers, exists := w.Header()[corsHeader]; !exists || len(headers) < 2 {
+				continue
+			}
+
+			logger.Warn().Msgf(`Multiple entries detected for header "%s". Only one should be set: you should enable CORS handling in the CLI only if the application does not handle them.`, corsHeader)
+		}
 	})
 }
--- a/local/http/cors.go
+++ b/local/http/cors.go
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2021-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package http
+
+import (
+	"net/http"
+)
+
+func corsWrapper(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", "*")
+		w.Header().Set("Access-Control-Allow-Methods", "*")
+		w.Header().Set("Access-Control-Allow-Headers", "*")
+
+		h.ServeHTTP(w, r)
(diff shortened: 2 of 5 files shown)


local/http/http.go
```diff
@@ -212,6 +213,10 @@ func (s *Server) ProxyHandler(w http.ResponseWriter, r *http.Request) {
 
 // Handler handles HTTP requests
 func (s *Server) Handler(w http.ResponseWriter, r *http.Request) {
+	if s.AllowCORS {
+		w.Header().Add("Access-Control-Allow-Origin", "*")
```
tucksaun, line 217: to make it actually useful on my side I had to add more headers:
```
if origin := r.Header.Get("origin"); origin != "" {
                w.Header().Set("Access-Control-Allow-Origin", origin)
                w.Header().Set("Access-Control-Allow-Methods", "GET, POST, OPTIONS")
                w.Header().Set("Access-Control-Allow-Headers", "DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,If-None-Match")
                w.Header().Set("Access-Control-Max-Age", "3600")
                w.Header().Set("Access-Control-Allow-Credentials", "true")
            }
```

this is actually why I didn't submit any PR yet it's because I was not satisfied with this and didn't see a clean why to do this.
anyone got some feedbacks?

Kocal: Maybe this could be configurable from the `.symfony.local.yaml` or another file?

tucksaun:
> Is there a way to add automated tests for that?

you should be able to write a functional test as you would in Go but I believe we don't have a test case available to get you inspired

local/http/http.go
tucksaun: Sorry for the delay @Kocal 😔
Actually, I didn't use the CLI on this project and probably won't for a while.
So I guess we can move forward!

tucksaun:
@Kocal did you get a change to have a look for a test?

Kocal:
@tucksaun not at all

tucksaun:
@Kocal I guess we are good, it's not worst than the actual situation :)

local/http/http.go
tucksaun, line 217: I'm wondering if we should not add default CORS values (see what Gorillw used to have: https://github.com/gorilla/handlers/blob/546854cf1d61a016260b224e2526a29ff5b5c5ae/cors.go#L28-L32)

tucksaun: We should also be careful not to override header sent by the upstream app

tucksaun (commented):
Should we check for preflight requests as well?

sarim:
I use `nelmio/cors-bundle` for this and it works perfectly for api-platform for a long time now. `composer require nelmio/cors-bundle` and `CORS_ALLOW_ORIGIN=*` would accomplish the same with better control. 

 What is the use case for this feature? For all application request I'd rather let the application handle the headers. This would be useful only for static files? Though we normally don't load static files over XHR/fetch. I think this should only be added to headers if symfony-cli is serving static file. If servering a request via php, let php handle the headers.

tucksaun:
Serving static files is one use case indeed.
Usually you don't load them via XHR except when this is a JS framework loading some templates (or similar use cases).

But you can also imagine running a JS development server alongside the CLI and both can be on different domain/port and this causes issues. But only in dev, so you might not want to (or not allowed to) introduce nelmio cors.

broncha:
I just had a need for this as I am serving a static json file from public folder and its not handled by `nelmio/cors-bundle`

Kocal:
PR rebased and conflicts fixed, I will see to add a test if I can

local/http/http.go
tucksaun: no this would not work because at this stage the header will never be already defined because the upstream server has not been contacted yet (contacting the upstream server is done in `s.Callback`)

Kocal:
I think I will need some help to finalize this PR (fix the "add header" logic and tests). 
If someone wants to help me (or re-open a new open if it's easier).. thanks :)

tucksaun:
> I think I will need some help to finalize this PR (fix the "add header" logic and tests). If someone wants to help me (or re-open a new open if it's easier).. thanks :)

If you happen to be at the Forum PHP in 15 days we can catch-up there and have a look :)

commands/local_server_start.go
```diff
@@ -52,6 +52,7 @@ import (
 
 var localWebServerProdWarningMsg = "The local web server is optimized for local development and MUST never be used in a production setup."
 var localWebServerTlsKeyLogWarningMsg = "Logging TLS master key is enabled. It means TLS connections between the client and this server will be INSECURE. This is NOT recommended unless you are debugging the connections."
+var localWebServerAllowCorsLogWarningMsg = "Cross-origin resource sharing (CORS) is enabled for every requests.\nYou may want to use https://github.com/nelmio/NelmioCorsBundle to have more control on HTTP headers."
```
sarim, line 55: Maybe add another sentence informing user that if they rely on this behavior to make their application work, it won't work in production server oob.

tucksaun:
@Kocal after a week thinking about our peer-programing session, I stand by what I was saying before leaving the Forum PHP: I believe we can accept your PR where the CLI prepends the headers and ignore the case where the application also add some headers. If this happens then I think this is the user’s responsibility to decide between disabling CORS headers addition in their app or in the CLI.

what we might want to do, just to be clear for the user, is add a warning if we detect such scenario but that’s it.

Kocal:
Hey @tucksaun, sorry for the late reply, I've been quite busy those last weeks :) 

That's fine to me, a warning should be enough, I'll check it out as soon as I can!

PS: it was a real pleasure to meet you at the PHP Forum :D

tucksaun:
Nice! Let me know if you need some help to finalize it.

PS: shared pleasure :)

rustymcfly:
Please, Merge this and create a release! This is the Feature i am waiting for :)

Kocal:
Ah! My bad, I didn't find the time to complete this PR, I just needed to commit and push :) 

It's should be ok now, cc @tucksaun :)

Kocal (commented):
Thanks @tucksaun for the last modifications 😍

tucksaun (approved):

commands/local_server_start.go
fabpot, line 55:
```suggestion
var localWebServerAllowCorsLogWarningMsg = "Cross-origin resource sharing (CORS) is enabled for all requests.\nYou may want to use https://github.com/nelmio/NelmioCorsBundle to have better control over HTTP headers."
```

fabpot, line 55:
```suggestion
var localWebServerAllowsCORSLogWarningMsg = "Cross-origin resource sharing (CORS) is enabled for every requests.\nYou may want to use https://github.com/nelmio/NelmioCorsBundle to have more control on HTTP headers."
```

local/http/cors.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+
+		h.ServeHTTP(w, r)
+
+		for _, corsHeader := range corsHeaders {
+			if headers, exists := w.Header()[corsHeader]; !exists || len(headers) < 2 {
+				continue
+			}
+
+			logger.Warn().Msgf(`Multiple entries detected for header "%s". Only one should be set: you should enable the CORS handling in the CLI only if you are application does not handle them.`, corsHeader)
```
fabpot, line 43:
```suggestion
			logger.Warn().Msgf(`Multiple entries detected for header "%s". Only one should be set: you should enable CORS handling in the CLI only if the application does not handle them.`, corsHeader)
```

fabpot (changes requested):

commands/local_server_start.go
```diff
@@ -188,6 +190,10 @@ var localServerStartCmd = &console.Command{
 			ui.Warning(localWebServerTlsKeyLogWarningMsg)
 		}
 
+		if config.AllowCORS {
+			ui.Warning(localWebServerAllowCorsLogWarningMsg)
```
tucksaun, line 194:
```suggestion
			ui.Warning(localWebServerAllowsCORSLogWarningMsg)
```

commands/local_server_start.go
tucksaun: thanks. updated

tucksaun: done

local/http/cors.go
tucksaun: thank you. updated

tucksaun (approved):

Kocal:
We dit it 🙌🏻

tucksaun:
Thank you @Kocal for keeping up the effort for a year and half ! 🥳

Kocal:
Thank you for our pair-programming sessions and your patience 🙏🏻

PR #295, opened 2023-03-22
Extract PostgreSQL, MySQL and MariaDB from Docker images in use, fix #108

### LINKED ISSUE #108: Append serverVersion to DATABASE_URL for local server
When using the local server, I noticed that the `serverVersion` parameter is not appended to the `DATABASE_URL` variable.

This would workaround an issue I encountered while running a functional test suite which failed because of too many opened connections.

I found that when I use `dmaicher/doctrine-test-bundle`, the way the connection is created open a connection to the database to get the server version (https://github.com/dmaicher/doctrine-test-bundle/blob/master/src/DAMA/DoctrineTestBundle/Doctrine/DBAL/StaticConnectionFactory.php#L39).

If I set the server version in `doctrine.yaml`, I can workaround the issue but I would like to avoid such situation as I may run my app with different PostgreSQL versions.

It would be a nice addition to append the `serverVersion` parameter to `DATABASE_URL` when using the local server.

tucksaun:


envs/docker.go
```diff
@@ -292,13 +296,16 @@ func (l *Local) dockerServiceToRelationship(client *docker.Client, container typ
...
 			path := ""
+			version := ""
 			for _, env := range c.Config.Env {
 				if strings.HasPrefix(env, "POSTGRES_USER") {
 					username = getEnvValue(env, "POSTGRES_USER")
 				} else if strings.HasPrefix(env, "POSTGRES_PASSWORD") {
 					password = getEnvValue(env, "POSTGRES_PASSWORD")
 				} else if strings.HasPrefix(env, "POSTGRES_DB") {
 					path = getEnvValue(env, "POSTGRES_DB")
+				} else if strings.HasPrefix(env, "PG_VERSION") {
```
fabpot, line 307: Is it something "standard"?

tucksaun: This is the value used in the official postgresql image, see https://github.com/docker-library/postgres/blob/3db64a0e20b456a554b9da8791704cb3e54af7f2/Dockerfile-alpine.template#L20

tucksaun: But nothing more official than this

PR #299, opened 2023-03-24
feat: remove slash on the elasticsearch url

shyim:
Right now the URL is like `http://127.0.0.1:49155/`

This causes errors like:
```
BadRequest400Exception {"error":"no handler found for uri [/_search] and method [POST]"}
```
because on OpenSearch Client it will append a double slash to all urls.

So I have removed the trailing slash


tucksaun:
This was added on purpose: see #155 and #154

shyim:
I will adjust it in our application. But for me it's still a issue in symfony-cli and platform.sh is for me not "the standard way" to do it :)

Issue #298, opened 2023-03-24
Feature request: add environment variable to skip new version check


yesworld:
Q: How do I disable the INFO of a new version?

```shell
$ symfony 

 INFO  A new Symfony CLI version is available (5.5.2, currently running 5.5.1).

       If you installed the Symfony CLI via a package manager, updates are going to be automatic.
       If not, upgrade by downloading the new version at https://github.com/symfony-cli/symfony-cli/releases
       And replace the current binary (symfony) by the new one.

```

fabpot:
By upgrading. There is no other way right now.

yesworld:
Just always update and that's it?
Or will there be some env parameter  where you can turn off the notification of this?

fabpot:
Note that it's disabled in non-interactive mode.
For interactive sessions, I think we need to warn our users of new versions and make sure they update at their earliest convenience.

Why would you like to disable it and not upgrade to the latest version?

yesworld:
I usually don't refresh every time if everything works fine. But unfortunately for me, as a perfectionist. It will be distracting every time I start the ```$ symfony server:start```.

Wouldn't it be better to make it informative for minor versions?

P.s. If I'm the only one, then I suggest closing Issue. I don't want to waste your time.

xorik:
I was also looking for this feature.

My reasoning:
Those messages distract me from my current task so I lose my focus, forget what I was doing, etc. It might be annoying. Plus in 99% this update won't change anything for me.

Another reason is that it's not common for terminal tools ask for updates. Imagine you run `git` and it ask you to update itself.

I agree disabling updates is terrible idea, but maybe make it less frequent, e.g. promote major releases, or suggest only if current version is 3-6 months old.

cafferata:
> Why would you like to disable it and not upgrade to the latest version?

Because it doesn't add anything when I update automatically by [homebrew-autoupdate](https://github.com/Homebrew/homebrew-autoupdate).

fabpot:
I like the reasoning here: https://github.com/symfony-cli/symfony-cli/pull/301#issuecomment-1488924755
At least, I don't think it makes sense for me to spend more time on this issue.

fabpot:
#309 limits the version check once per day, I think this is a good compromise.

PR #301, opened 2023-03-29
Disable upgrade message when installed via a package manager

fabpot:
refs #298

Not sure how to do it for apt or yum as they are using standard locations. Or maybe we should have a special build for package managers? That would make things much easier.

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -20,9 +20,12 @@
 package commands
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


tucksaun:
> Or maybe we should have a special build for package managers

That would be the way to go indeed.

But I would argue that the reasoning exposed in https://github.com/symfony-cli/symfony-cli/issues/298#issuecomment-1485496083 is wrong here: it's not because one uses a package manager that they monitor package updates and even less that they apply upgrades (otherwise security issues, fragmentation and maintenance windows would not be problems in our industry).

After all the warning happens (or at least it's supposed to happen) only once a day at the moment so it should not be too much of a trouble and if one sees the warning it means there is actually an upgrade to apply....
Maybe we can only do the warning once per version? Or just have a file similar to `.silence` that would make the binary not check updates at all?

But if we go the way of disabling the warning for packaged version then having the check is not even worth it anymore IMHO as it would only be for a minority of people as they are encouraged to use package managers anyway.

fabpot:
I think I agree with you 100% :)
Let's close.

PR #302, opened 2023-04-01
Allow user to ignore new version check

### LINKED ISSUE #298: Feature request: add environment variable to skip new version check

henrikac:
This PR attempts to fix #298 . It allows the user to skip the version check by setting `SYMFONY_IGNORE_NEW_VERSION=1`.

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -115,12 +115,14 @@ func InitAppFunc(c *console.Context) error {
 	if c.App.Channel == "stable" {
 		// do not run auto-update in the cloud, CI or background jobs
 		if !util.InCloud() && terminal.Stdin.IsInteractive() && !reexec.IsChild() {
-			debug := false
-			if os.Getenv("SC_DEBUG") == "1" {
-				debug = true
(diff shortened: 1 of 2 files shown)


henrikac:
I'm not sure if the environment variable should be prefixed with `SYMFONY_` or if there is a better prefix to use in this situation?

cpetit-sigma-fr:
Prefix with `SYMFONY_CLI_`?

henrikac:
> Prefix with SYMFONY_CLI_?

Is this the prefix used for the environment variables that is specific to the symfony-cli?

fabpot:
#309 limits the version check once per day, I think this is a good compromise.

PR #311, opened 2023-04-21
feat(chore): add golangci-lint

darkweak:
Here is the rebased PR with only needed changes @tucksaun 

* Only already checked errors are handled by the `errors.WithStack` wrapper
* Others are just ignored
* Only few linters are enabled at the moment but the list can be increased later in another PR

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -64,6 +64,11 @@ jobs:
                     echo '"go generate" changed some Go generated code, run "symfony cloud:self-update" then "go generate ./" locally and make a Pull Request with the changes'
                     git diff
                     exit 1
+            - 
+                name: Run golangci-lint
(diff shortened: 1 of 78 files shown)


tucksaun:
@darkweak did I miss something? the PR is now bigger than #245 😅

darkweak:
@tucksaun there are new returned errors now 😅

tucksaun:
This is not as simple as a search and replace as some are not necessary or some could change the semantic of the error by changing its type and thus could break check based on it which is particularly risky given the fact that we miss tests on those.

Regarding your question, given the size of this PR and my current free time that I can allow to OSS I double I will have time soon enough to review this PR in a single iteration quickly enough so that there's not conflict and no need for back and forth. That's why I invited you to submit smaller scoped contributions in your original PR that way we can more easily assess the changes and merged them without blocking the whole contribution.

PR #333, opened 2023-07-11
Merge $_ENV into $_SERVER earlier in phprouter file

kick-the-bucket:
Tried running a non-symfony application with the symfony-cli local server and was puzzled why it could not detect that it's running on https. Apparently `'https' => 'on'` was only set in `$_ENV` (that's done with `require $envFile;`), but not in `$_SERVER`, where PHP applications usually look for that parameter.

This was because the `$_SERVER = array_merge($_SERVER, $_ENV);` line was never executed, because the router exited a few lines before it with `return false;`.

Also fixed a few whitespace and indentation inconsistencies.

### DIFF
--- a/local/php/php_builtin_server.go
+++ b/local/php/php_builtin_server.go
@@ -37,22 +37,23 @@ var phprouter = []byte(`<?php
 
 // Workaround https://bugs.php.net/64566
 if (ini_get('auto_prepend_file') && !in_array(realpath(ini_get('auto_prepend_file')), get_included_files(), true)) {
-    require ini_get('auto_prepend_file');
+	require ini_get('auto_prepend_file');
 }
 
 if (isset($_SERVER['HTTP___SYMFONY_LOCAL_REQUEST_ID__'])) {
-if (file_exists($envFile = __FILE__.'-'.$_SERVER['HTTP___SYMFONY_LOCAL_REQUEST_ID__'].'-env')) {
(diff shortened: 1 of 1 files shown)


tucksaun:
PHP documents the presence of `HTTPS` within `$_SERVER` (see https://www.php.net/manual/en/reserved.variables.server.php), not `$_ENV`.
And I believe most of the literature online also references `$_SERVER`.

So I think we are within spec here, your app should probably be fixed instead.

tucksaun:
Also I don't quite understand how your application could even be reached if the router exited before the call to `array_merge` 🤔

kick-the-bucket:
After that `return false;` it just continues to my `index.php`, not really sure how that works, maybe the router file is set as an `auto_prepend_file` :thinking:

kick-the-bucket:
> PHP documents the presence of `HTTPS` within `$_SERVER` (see https://www.php.net/manual/en/reserved.variables.server.php), not `$_ENV`.
> And I believe most of the literature online also references `$_SERVER`.

The problem is that the code in the `$envFile` only sets `$_ENV['https'] = 'on'`, but nothing on the `$_SERVER` and that is why the merge is needed in the first place

kick-the-bucket:
@tucksaun I actually like your version, no point in setting the variables on `$_ENV` which are supposed to be in `$_SERVER` anyways :+1:

Issue #91, opened 2022-02-09
X Server crashes when starting server


dealer00:
I'm using Linux Mint 19.1 Cinnamon. Sometimes when I run `symfony serve` my X server crashes. Everything works fine after restart but all applications close immediately and you risk losing data.
Some time ago a had problem starting server because symfony says that port is currently used but it was not. Maybe there is problem with stopping server?

Logs from syslog:
`Feb  9 12:49:12 XXX at-spi-bus-launcher[2487]: XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"`
`Feb  9 12:49:12 XXX at-spi-bus-launcher[2487]:       after 31651 requests (31651 known processed) with 0 events remaining.`
`Feb  9 12:49:12 XXX xdg-desktop-por[3333]: xdg-desktop-portal-gtk: Fatal IO error 11 (Zasoby chwilowo niedostępne) on X server :0.`
`Feb  9 12:49:12 XXX systemd[2377]: xdg-desktop-portal-gtk.service: Main process exited, code=exited, status=1/FAILURE`
`Feb  9 12:49:12 XXX systemd[2377]: xdg-desktop-portal-gtk.service: Failed with result 'exit-code'.`
`Feb  9 12:49:12 XXX gnome-terminal-[10281]: gnome-terminal-server: Fatal IO error 11 (Zasoby chwilowo niedostępne) on X server :0.`
`Feb  9 12:49:12 XXX systemd[2377]: gnome-terminal-server.service: Main process exited, code=exited, status=1/FAILURE`
`Feb  9 12:49:12 XXX systemd[2377]: gnome-terminal-server.service: Failed with result 'exit-code'.`
`Feb  9 12:49:13 XXX lightdm[1282]: Error: can't open /lib/modules/4.15.0-167-generic/updates/dkms`
`Feb  9 12:49:13 XXX lightdm[1282]: Error: can't open /lib/modules/4.15.0-167-generic/updates/dkms`
`Feb  9 12:49:13 XXX systemd[1]: nvidia-persistenced.service: Unit not needed anymore. Stopping.`
`Feb  9 12:49:13 XXX systemd[1]: Requested transaction contradicts existing jobs: Transaction is destructive.`
`Feb  9 12:49:13 XXX systemd[1]: nvidia-persistenced.service: Failed to enqueue stop job, ignoring: Transaction is destructive.`
[...]

PR #338, opened 2023-08-11
Prevent potential panic when proxy fails connecting to local server

tucksaun:
Refs https://github.com/symfony-cli/symfony-cli/issues/91#issuecomment-1650245926


henrikac:
The reason why you're not checking if `targetSiteCon != nil` the other places where `targetSiteCon.Close()` is called is because you're sure that `targetSiteCon != nil`? or would it be possible for it to be `== nil` on e.g. line 116?

tucksaun:
I don't check if `targetSitCon` is nil in other places because the reasons that would lead to have it nil would also have make us got an error at line 95, so we would not reach those other places.

fabpot (approved):

PR #347, opened 2023-09-26
Fix compatibility with Doctrine DBAL 4

fabpot:


javiereguiluz:
According to @stof, support for this format was introduced in DBAL 2.10.0 ... so 210.x and 3.x and 4.x should support it.

Kocal:
Alright thanks!

PR #364, opened 2023-10-24
Add support for compose.yaml

fabpot:


dunglas (approved):

dunglas:
Ideally, Symfony CLI should also read the `COMPOSE_FILE` env var, because Flex supports this. https://docs.docker.com/compose/environment-variables/envvars/#compose_file

fabpot:
> Ideally, Symfony CLI should also read the `COMPOSE_FILE` env var, because Flex supports this. https://docs.docker.com/compose/environment-variables/envvars/#compose_file

That's already the case a few lines above :)

PR #368, opened 2023-11-02
Expose the CLI version in the WDT

### LINKED ISSUE #365: Display the CLI version in the inject web debug toolbar item
Currently, the item injected in the toolbar shows the version of the server being used and some other info. It would help debugging things like https://github.com/symfony/symfony/issues/52320 if it also showed the version of the symfony-cli.

tucksaun:
fix #365 

Screenshot 2023-11-02 at 17 29 39

while working on this I noticed the icon in the WDT was not optimal on small screens so I took the opportunity to fix classes to make it displayed as expected:
Screenshot 2023-11-02 at 17 29 20

### DIFF
--- a/local/php/toolbar.go
+++ b/local/php/toolbar.go
@@ -129,7 +129,10 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
 	<div class="sf-toolbar-info" style="left: 0px;">
 		<div class="sf-toolbar-info-group">
 			<div class="sf-toolbar-info-piece">
-				<b>Server</b>` + p.Version.ServerTypeName() + ` ` + p.Version.Version + `
+				<b>Server</b>` + p.appVersion + `
+			</div>
+			<div class="sf-toolbar-info-piece">
+				<b>PHP</b>` + p.Version.ServerTypeName() + ` ` + p.Version.Version + `
 			</div>
 			<div class="sf-toolbar-info-piece">
 				<b>Tunnel</b>` + tunnel + `
--- a/local/php/toolbar.go
+++ b/local/php/toolbar.go
@@ -129,7 +129,7 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
 	<div class="sf-toolbar-info" style="left: 0px;">
 		<div class="sf-toolbar-info-group">
 			<div class="sf-toolbar-info-piece">
-				<b>Server</b>` + p.appVersion + `
+				<b>Symfony CLI</b>` + p.appVersion + `
 			</div>
 			<div class="sf-toolbar-info-piece">
 				<b>PHP</b>` + p.Version.ServerTypeName() + ` ` + p.Version.Version + `
--- a/local/php/php_server.go
+++ b/local/php/php_server.go
@@ -51,6 +51,7 @@ import (
 type Server struct {
 	Version      *phpstore.Version
 	logger       zerolog.Logger
+	appVersion   string
 	homeDir      string
 	projectDir   string
 	documentRoot string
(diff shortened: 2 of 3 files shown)


tucksaun:
@javiereguiluz can you confirm the class fix is the right one?

stof:
@tucksaun you should put the `fix #365` in the PR description rather than the PR title as cross-references don't work in the title.

local/php/toolbar.go
```diff
@@ -116,19 +116,22 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
 		return body, errors.WithStack(err)
 	}
 	content := []byte(`
-<div class="sf-cli sf-toolbar-block sf-toolbar-block-sf-cli ` + logoBg + ` sf-toolbar-block-right">
+<div class="sf-cli sf-toolbar-block-config sf-toolbar-block sf-toolbar-block-sf-cli ` + logoBg + ` sf-toolbar-block-right">
```
stof, line 119: I'm quite sure `sf-toolbar-block-config` is expected to be used only by the config panel.

local/php/toolbar.go
```diff
@@ -116,19 +116,22 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
...
 			</svg>
 		</span>
-		<span class="sf-toolbar-value">Server</span>
+		<span class="sf-toolbar-value">Server ` + p.appVersion + `</span>
```
stof, line 127: I would keep the app version only in the info, not in the icon section, to take less space (the CLI version is useful, but not important enough to take space inline instead of requiring to hover to see the info)

tucksaun:
> @tucksaun you should put the `fix #365` in the PR description rather than the PR title as cross-references don't work in the title.

done

local/php/toolbar.go
stof: I'm quite sure the bad display in small screen is a bug in WebprofilerBundle where one of the CSS selector uses the `sf-cli` class instead of `sf-toolbar-block-sf-cli`

javiereguiluz:
Yes, the CSS class looks correct to me. Thanks!

local/php/toolbar.go
tucksaun: > I'm quite sure `sf-toolbar-block-config` is expected to be used only by the config panel.

that's what I thought first. But then I came to the conslusion that is this a class instead of an id that would probably be supported 🙂

> I'm quite sure the bad display in small screen is a bug in WebprofilerBundle where one of the CSS selector uses the sf-cli class instead of sf-toolbar-block-sf-cli

TBH I didn't expect the WebProfilerBundle to include any specific CSS for the CLI so I was surprised to see `sf-toolbar-block-sf-cli` in there 😮 
Then I guess I will open a PR on symfony/symfony ;)

tucksaun: NVM it has been fixed in https://github.com/symfony/symfony/pull/48961

tucksaun: reverted

local/php/toolbar.go
tucksaun: fine with me. updated

local/php/toolbar.go
```diff
@@ -129,7 +129,10 @@ func (p *Server) tweakToolbar(body io.ReadCloser, env map[string]string) (io.Rea
 	<div class="sf-toolbar-info" style="left: 0px;">
 		<div class="sf-toolbar-info-group">
 			<div class="sf-toolbar-info-piece">
-				<b>Server</b>` + p.Version.ServerTypeName() + ` ` + p.Version.Version + `
+				<b>Server</b>` + p.appVersion + `
```
fabpot, line 132:
```suggestion
				<b>Symfony CLI</b>` + p.appVersion + `
```

fabpot (changes requested):

PR #370, opened 2023-11-02
Add support for a `docker_compose` worker and prioritize its startup

### LINKED ISSUE #240: Allow configuring Docker compose to automatically start with server
Hi,

I think it would be nice to have an option (in .symfony.local.yaml?) to specify that it should also run Docker compose automatically when running the server.

Maybe also allowing to define the Docker compose files to use?
Like this:

```yaml
docker_compose:
    - docker-compose.yaml
    - docker-compose.override.yaml
```

So when running `server:start` it would run `docker compose up`? And if daemon is enabled with adding the `-d` option? (so when running `server:stop` it should also run `docker compose stop` if daemon was enabled)

tucksaun:
It should fix #240.
based on #308

The idea is that when this specific worker is configured it will block the start up of other workers.

Also to be "sure" services are up we do the startup in a two-step process:
1.  first we start `docker compose up -d` which will return only once services are up
2. then we:
   1. change the worker to start `docker compose up` instead, which will block until we stop the server and will stop the containers when being stopped
   2. unblock the other workers

### DIFF
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -29,6 +29,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"sync"
 	"syscall"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


tucksaun:
Not entirely ready yet: I encounter a bunch of "no such process" error when exiting but I'm not sure where they come from

tucksaun:
Actually, I get the "no such process" on `main` as well so I think we should be good for a review once #308 is merged :)

fabpot:
Can be rebased

tucksaun:
@fabpot rebased

fabpot:
Thank you @tucksaun

tucksaun:
@stof the PHP server determines the variables on every request so this should work out-of-the box

maxhelias:
Thans @tucksaun for this awesome feature 👏

PR #371, opened 2023-11-04
Add support for PHP streaming

### LINKED ISSUE #330: Support for streaming
Hi!

I'm experimenting with a dev-only implementation of HMR for AssetMapper where a Symfony endpoint "streams" information about changes made to CSS files. However, like #67, I don't believe the `symfony` binary server supports streams: the same code that streams with `php -S` does not stream using `symfony serve`.

Is something that would be possible to add? Without it, in practice, for HMR, I think we'd need to require the user to install Mercure, which would be a much higher hurdle.

Thanks!

tucksaun:
Fix #330 

Add support for PHP streaming.
On PHP builtin server side, the only change required was to move from buffering the response to golang's proxy `ModifyResponse`

On the CGI/FPM side this was a bit more tricky initially, but in the end because we deal with an HTTP request and a response, we can actually wrap the CGI communication within a Golang HTTP proxy context which make it work out of the box.

~Remaining feature to restore: X-Send-file~

ping @weaverryan

### DIFF
--- a/local/php/xsendfile.go
+++ b/local/php/xsendfile.go
@@ -8,7 +8,9 @@ import (
 func (p *Server) processXSendFile(resp *http.Response) (error, bool) {
 	// X-SendFile
 	sendFilename := resp.Header.Get("X-SendFile")
-	if _, err := os.Stat(sendFilename); sendFilename == "" || err != nil {
+	if sendFilename == "" {
+		return nil, false
+	} else if _, err := os.Stat(sendFilename); err != nil {
 		return nil, false
 	}
 
--- a/local/php/xsendfile.go
+++ b/local/php/xsendfile.go
@@ -1,0 +1,21 @@
+package php
+
+import (
+	"net/http"
+	"os"
+)
+
+func (p *Server) processXSendFile(resp *http.Response) (error, bool) {
+	// X-SendFile
+	sendFilename := resp.Header.Get("X-SendFile")
+	if _, err := os.Stat(sendFilename); sendFilename == "" || err != nil {
+		return nil, false
+	}
+
+	req := resp.Request
+	w := req.Context().Value(responseWriterContextKey).(http.ResponseWriter)
+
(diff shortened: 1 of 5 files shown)


tucksaun:
@weaverryan you should be able to test it locally using CI artifacts available here: https://github.com/symfony-cli/symfony-cli/suites/17905324339/artifacts/1028487092. let me know how it works for you!

fabpot:
Thank you so much for working on this one!

tucksaun:
X-Sendfile support has been restored.

I also had to tweak the WDT processing because now that we do streaming we need to preserve the ability to close the original body, this does not affect too much PHP builtin server or FPM but without this the CGI server gets blocked after first request serving the WDT.

And finally I pushed some cleanup so this should be ready for review @fabpot

local/php/xsendfile.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"net/http"
+	"os"
+)
+
+func (p *Server) processXSendFile(resp *http.Response) (error, bool) {
+	// X-SendFile
+	sendFilename := resp.Header.Get("X-SendFile")
+	if _, err := os.Stat(sendFilename); sendFilename == "" || err != nil {
```
fabpot, line 11: no need to stat is the file is empty

fabpot (commented):
I haven't tested it, but looks good to me.

local/php/xsendfile.go
tucksaun: I kept this part as-is but it's true this is the occasion to improve it! Done

tucksaun:
(BTW, the refactoring that happened means that we now support X-Sendfile with PHP builtin server as well)

fabpot:
Thank you @tucksaun

tucksaun:
NB for anyone willing to make it work: PHP has to either send a `Content-Type` set to `text/event-stream` or a `Content-Length` set to `-1`

PR #372, opened 2023-11-04
Use the original inotify repository

fabpot:
/cc @tucksaun


tucksaun:
didn't we pick the fork because it was more maintained or was fixing edge cases?

fabpot:
The fork is not maintained (last commit was in 2021) and it currently throws a warning on MacOS. The original repository is maintained and when quickly comparing the 2 repositories, it looks like everything from the fork is now in the main one.

tucksaun:
ok thanks for the update 👍

PR #378, opened 2023-11-07
Also consider 27018 and 27019 ports for MongoDB service discovery

### LINKED ISSUE #377: Missing environment variables
I have the following simple docker-compose file:
```yaml
version: "3.9"

services:
    mongodb_primary:
        image: mongo:7-jammy
        environment:
            MONGO_INITDB_ROOT_USERNAME: primary
            MONGO_INITDB_ROOT_PASSWORD: password
        ports: [ '27017' ]

    mongodb_secondary:
        image: mongo:7-jammy
        environment:
            MONGO_INITDB_ROOT_USERNAME: secondary
            MONGO_INITDB_ROOT_PASSWORD: password
        ports: [ '27018' ]
```

and the result is:
```terminal
$ symfony var:export --multiline

export MONGODB_PRIMARY_DATABASE=
export MONGODB_PRIMARY_DB=
export MONGODB_PRIMARY_HOST=127.0.0.1
export MONGODB_PRIMARY_NAME=
export MONGODB_PRIMARY_PASSWORD=password
export MONGODB_PRIMARY_PORT=63663
export MONGODB_PRIMARY_SCHEME=mongodb
export MONGODB_PRIMARY_SERVER=mongodb://127.0.0.1:63663
export MONGODB_PRIMARY_URL=mongodb://primary:password@127.0.0.1:63663/?authSource=
export MONGODB_PRIMARY_USER=primary
export MONGODB_PRIMARY_USERNAME=primary
export MONGODB_SECONDARY_HOST=127.0.0.1
export MONGODB_SECONDARY_IP=127.0.0.1
export MONGODB_SECONDARY_PORT=63664
export SYMFONY_DOCKER_ENV=1
export SYMFONY_TUNNEL=
export SYMFONY_TUNNEL_BRAND=Platform.sh
export SYMFONY_TUNNEL_ENV=
```

However, as you can see, a lot of `MONGODB_SECONDARY_*` environment variables are missing.

The version I am using:
`Symfony CLI version 5.6.2 (c) 2021-2023 Fabien Potencier #StandWithUkraine (2023-10-25T14:07:11Z - stable)`

tucksaun:
Fix #377

### DIFF
--- a/envs/docker.go
+++ b/envs/docker.go
@@ -397,7 +397,7 @@ func (l *Local) dockerServiceToRelationship(client *docker.Client, container typ
 				"scheme": "http",
 			}
 			return rels
-		} else if p.PrivatePort == 27017 {
+		} else if p.PrivatePort == 27017 || p.PrivatePort == 27018 || p.PrivatePort == 27019 {
 			username := ""
 			password := ""
 			path := ""


OskarStark:
While reading the fix, is it really the best approach to map this to the used ports? I mean if I need to use 3 MongoDB services (and we are talking about dev and testing mostly) I would run into the same issue.

What about checking some combination of port and/or image name?

Or am I able to set the type manually via options in the `compose.yaml` to tell SymfonyCLI "Hey, this is a MongoDB"?

alcaeus:
> While reading the fix, is it really the best approach to detect this by the used ports?

I'd argue "no", but it's usually the most concise way. While 27017 is the default port used for MongoDB on a local machine for a standalone server, this isn't the recommended setup for various reasons. While there's a certain convention to use the following ports for more servers (e.g. when setting up a replica set with multiple members on a single machine), I don't think it's feasible to start checking for all possible ports, especially since this gets quite complicated when you start involving sharded clusters.

I'm not too familiar with what exactly the purpose of this is, but I agree that there should either be a more specific check for MongoDB (e.g. by checking image names), or providing a method for people to specify what service to expose. That said, a compromise would be to cover the typical use cases:
* A standalone server running on port 27017
* A replica set of (up to) three members running on ports 27017, 27018, 27019
* A sharded cluster, where the mongos application listens on port 27017 (which is what the application should connect to).

Once we go beyond those setups, it becomes increasingly difficult to predict what the user's intention is, so I wouldn't bother with them.

fabpot:
As this is only for local setups, I would go with the typical use cases only, so the 3 mentioned ports.

tucksaun:
> What about checking some combination of port and/or image name?

Image name does not work with setups that build custom images :/
To me, the only viable alternative is to check for image labels as they would be preserved if the image is derived from another one.
But if I remember correctly, last time I checked this was not really convenient with the current API 😕  and not every official images use them 😞 

> While reading the fix, is it really the best approach to detect this by the used ports?

Indeed maybe not. But please note we check for the private port, not the one exposed on the host. The private port is the one the service is listening to inside the container. And, in the container world, it is less common to change it because we usually do the remap at the host level.

tucksaun:
> As this is only for local setups, I would go with the typical use cases only, so the 3 mentioned ports.

Third port added

PR #384, opened 2023-11-23
Add --api flag to local:new command to create a new API project

pierredup:
This will allow creating a new API project when using the `--api` flag, which will add the api pack


fabpot:
Thank you

Issue #349, opened 2023-10-04
Local server does not exit cleanly with CTRL+C


bakugo:
Using symfony-cli 5.5.10 on Windows 10.

Whenever I start the local server in the foreground with `symfony serve` and later attempt to stop it with CTRL+C, it *always* prints one of the following errors in red after "Shutting down!":

>TerminateProcess: Access is denied.

> remove C:\\Users\\User\\.symfony5\\log\\141412775c62bcc436990ef52288172b438b0562.log: The process cannot access the file because it is being used by another process.

It seems to stop properly despite these errors.

klodoma:
+1

PR #394, opened 2023-12-10
Rework workers lifecycle management to make them stop cleanly

### LINKED ISSUE #390: Command server:stop doesn't stop docker_compose worker if server was started with --deamon
To reproduce

- have in `.symfony.local.yaml`
    ```yaml
    workers:
        docker_compose: ~
    ```
- run `symfony server:start --daemon`
- run `docker compose ps` to see containers running
- run `symfony server:stop`
- run `docker compose ps` again to see containers still running

is this an expected behavior?

tucksaun:
The workers will already receive the SIGINT signal because Ctrl+C is sent to the whole process group in which they are. This means that in the `server:start` command we only need to disable worker restart and wait for them to finish before exiting. In the `server:stop` command, this means we have to first signal the monitoring process (but without stopping it).

Fix #390

### DIFF
--- a/commands/local_proxy_start.go
+++ b/commands/local_proxy_start.go
@@ -167,7 +167,7 @@ var localProxyStartCmd = &console.Command{
 		shutdownCh := make(chan bool, 1)
 		go func() {
 			sigsCh := make(chan os.Signal, 1)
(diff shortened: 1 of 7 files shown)


tucksaun:
@fabpot would you have a Windows machine available to test this out?

tucksaun:
Might also improve things for #349

fabpot:
This PR does improve the situation on Windows.

PR #395, opened 2023-12-14
Fix speed issue

fabpot:


local/platformsh/applications.go
```diff
@@ -198,7 +203,7 @@ func findAppConfigFiles(brand CloudBrand, dir string) []string {
 			}
 
 			// don't go too deep down the tree
-			if len(strings.Split(path[rootDirectoryLen:], separator)) > 5 {
+			if len(strings.Split(path[rootDirectoryLen:], separator)) > 3 {
```
fabpot, line 206: @tucksaun I don't think going that deep is needed. WDYT?

tucksaun: Depends how deep you want to support multi app projects. 5 might sounds a bit too high

Issue #159, opened 2022-07-19
New command to change tld for symfony proxy


MrYamous:
Hi,

When we use Symfony local proxy, default tld is ".wip"
[Documentation](https://symfony.com/doc/current/setup/symfony_server.html#defining-the-local-domain) explain how to change it manually, but what would you think about adding a command to do it ? 

It's not a major improvement but I guess it can still be helpful and maybe more convenient to use

Nihilus118:
Hi @tucksaun,

please take a look at the branch I just created. I've implemented a new command to change the top-level domain (TLD) of the local proxy. Additionally, I've made modifications to the existing command `local:proxy:domain:attach` to support an argument for specifying a TLD when attaching domains.

I'm uncertain whether this approach is the most optimal or if it aligns with your considerations for backward compatibility.

tucksaun:
@Nihilus118 thank you for working on this.

a PR would be easier for @fabpot and I to review the code (and to comment).

Nihilus118:
Hi @tucksaun, I created a PR as requested. The pipeline also passed successfully.

PR #396, opened 2023-12-16
Add command to change the tld of local proxy (issue #159)

Nihilus118:
PR for issue #159

### DIFF
--- a/commands/local_proxy_change_tld.go
+++ b/commands/local_proxy_change_tld.go
@@ -52,7 +52,7 @@ var localProxyTLD = &console.Command{
 			return err
 		}
 
-		terminal.Printfln("<info>The proxy is now configured with the following tld: %s</>", config.TLD)
+		terminal.Printfln("<info>The proxy is now configured with the following TLD: %s</>", config.TLD)
 		return nil
 	},
 }
--- a/commands/local_proxy_change_tld.go
+++ b/commands/local_proxy_change_tld.go
@@ -45,8 +45,8 @@ var localProxyTLD = &console.Command{
 		}
 
 		config.TLD = c.Args().Get("tld")
-		if !regexp.MustCompile(`^[a-z0-9-]{2,63}$`).MatchString(config.TLD) {
-			return fmt.Errorf("the TLD must only contain lowercase letters, numbers and hyphens")
+		if !regexp.MustCompile(`^[a-z]{1,63}$`).MatchString(config.TLD) {
+			return fmt.Errorf("the TLD must only contain lowercase letters")
 		}
 		if err = config.Save(); err != nil {
 			return err
--- a/commands/local_proxy_change_tld.go
+++ b/commands/local_proxy_change_tld.go
@@ -31,7 +31,7 @@ import (
 
 var localProxyTLD = &console.Command{
 	Category: "local",
-	Name:     "proxy:change:tld",
+	Name:     "proxy:tld",
 	Aliases:  []*console.Alias{{Name: "proxy:change:tld"}},
 	Usage:    "Change the TLD for the proxy",
 	Args: []*console.Arg{
--- a/commands/local_proxy_change_tld.go
+++ b/commands/local_proxy_change_tld.go
@@ -20,53 +20,36 @@
 package commands
 
 import (
-	"strings"
-
 	"github.com/symfony-cli/console"
 	"github.com/symfony-cli/symfony-cli/local/proxy"
 	"github.com/symfony-cli/symfony-cli/util"
 	"github.com/symfony-cli/terminal"
 )
 
-var localProxyAttachDomainCmd = &console.Command{
+var localProxyTLD = &console.Command{
 	Category: "local",
-	Name:     "proxy:domain:attach",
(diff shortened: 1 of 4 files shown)


commands/local_proxy_domain_attach.go
```diff
@@ -48,7 +51,16 @@ var localProxyAttachDomainCmd = &console.Command{
...
+
+		var domains []string
+		if strings.HasPrefix(c.Args().Slice()[0], ".") {
+			config.TLD = strings.TrimPrefix(c.Args().Slice()[0], ".")
+			domains = c.Args().Slice()[1:]
+		} else {
+			domains = c.Args().Slice()
+		}
+
+		if err := config.AddDirDomains(projectDir, domains); err != nil {
```
tucksaun, line 63: I don't think this makes sense to allow attaching and changing the TLD at the same time: this cover a quite small use case and it could be confusing for the end-user letting them think this TLD is only for the specific project (which is not the case).

tucksaun (commented):
Thank you for this contribution.

Here are a couple of comments to move forward.

commands/local_proxy_domain_attach.go
Nihilus118: You are right. I reverted the file back to it's previous state.

commands/local_proxy_change_tld.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+var localProxyTLD = &console.Command{
+	Category: "local",
+	Name:     "proxy:change:tld",
+	Aliases:  []*console.Alias{{Name: "proxy:change:tld"}},
+	Usage:    "Change the TLD for the proxy",
+	Flags: []console.Flag{
+		dirFlag,
+	},
```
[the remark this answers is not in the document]
Nihilus118: Removed.

commands/local_proxy_change_tld.go
```diff
@@ -0,0 +1,55 @@
...
+		{Name: "tld", Optional: false, Description: "The TLD for the project proxy", Slice: false},
+	},
+	Action: func(c *console.Context) error {
+		homeDir := util.GetHomeDir()
+		config, err := proxy.Load(homeDir)
+		if err != nil {
+			return err
+		}
+
+		config.TLD = c.Args().Get("tld")
```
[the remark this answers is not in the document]
Nihilus118: What do you think about the regular expression `^[a-z0-9-]{2,63}$` for validating the tld before saving?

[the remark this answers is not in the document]
tucksaun: IIRC, the top level domain is only allowed by the RFC to be composed of 1 to 63 alpha chars (except for IDN but I'm not sure this makes sense to support them for now)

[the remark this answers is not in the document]
Nihilus118: I adjusted the regular expression accordingly.

commands/local_proxy_change_tld.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/symfony-cli/console"
+	"github.com/symfony-cli/symfony-cli/local/proxy"
+	"github.com/symfony-cli/symfony-cli/util"
+	"github.com/symfony-cli/terminal"
+)
+
+var localProxyTLD = &console.Command{
+	Category: "local",
+	Name:     "proxy:change:tld",
```
fabpot, line 34:
```suggestion
	Name:     "proxy:tld",
```

With no arguments, it would return the current TLD, and set it if an argument is provided.

commands/local_proxy_change_tld.go
```diff
@@ -0,0 +1,58 @@
...
+	Aliases:  []*console.Alias{{Name: "proxy:change:tld"}},
+	Usage:    "Change the TLD for the proxy",
```
fabpot, line 36:
```suggestion
	Usage:    "Display or change the TLD for the proxy",
```

commands/local_proxy_change_tld.go
```diff
@@ -0,0 +1,58 @@
...
+
+		config.TLD = c.Args().Get("tld")
+		if !regexp.MustCompile(`^[a-z]{1,63}$`).MatchString(config.TLD) {
+			return fmt.Errorf("the TLD must only contain lowercase letters")
+		}
+		if err = config.Save(); err != nil {
+			return err
+		}
+
+		terminal.Printfln("<info>The proxy is now configured with the following tld: %s</>", config.TLD)
```
fabpot, line 55:
```suggestion
		terminal.Printfln("<info>The proxy is now configured with the following TLD: %s</>", config.TLD)
```

fabpot (changes requested):
LGTM, I've just made a small suggestion, but maybe I'm overthinking it.

Nihilus118:
> LGTM, I've just made a small suggestion, but maybe I'm overthinking it.

You are right, this is a better description of the command. I adjusted the filename for consistent naming, added another alias, made the argument `tld` optional and wrapped the logic for adjusting the TLD in a call to `c.Args().Present()` to prevent this error which occured when no new TLD was provided:
image

fabpot (approved):
Thank you @Nihilus118

Issue #404, opened 2024-01-04
"symfony run" interactivity broken after v5.7.5


sarim:
"symfony run" interactivity broken after https://github.com/symfony-cli/symfony-cli/commit/ff1958fbb1f875372c8d654e37f4c7c29327e9f2

In this commit runner code was changed, the specific line is

https://github.com/symfony-cli/symfony-cli/blob/041d0074cfd2134840943aa623d7ee57bf65ed2b/local/runner_posix.go#L11-L16

Setting `Setpgid` to true loses stdin connection to terminal from the command, resulting in this.

```
$ symfony run psql
psql (15.5 (Ubuntu 15.5-1.pgdg22.04+1))
Type "help" for help.

```

It just hangs there. Anything typed in the terminal doesn't go through to the psql process. You can check by simple `symfony run cat`.

If I set Setpgid to false and compile it works. But that would obviously break the the original issue the referenced commit fixed. I think a proper solution would be [tcgetpgrp](https://man7.org/linux/man-pages/man3/tcgetpgrp.3.html), which we can make golang call if we also set `Foreground: true`

```go
	cmd.SysProcAttr = &syscall.SysProcAttr{
		// isolate each command in a new process group that we can cleanly send
		// signal to when we want to stop it
		Setpgid:    true,
		Foreground: true,
	}
```

This fixes interactive commands. We have to take care to set `Foreground: true` only when user is running symfony run something interactively, not for worker/daemons. So Runner.Mode needs to be passed to buildCmd function.

I'll send a PR with complete fix.

update; PR opened #405

sarim:
Reverted fornow (with #410 as well) as it doesn't work (at least on Macos where it hangs and ^C doesn't work anymore).

_Originally posted by @fabpot in https://github.com/symfony-cli/symfony-cli/issues/405#issuecomment-1900190365_

@fabpot can we open this again as the fix was reverted? Can you provide a bit more info about how it doesn't work? From my testing I managed to reproduce **the hangs and ^C doesn't work** only if `--watch` flag was also provided. So a command like
```
symfony run --watch=config sleep 3600
```
It ONLY breaks after a change is detected, which kinda makes sense because we were making the cmd process to have its own pgid, then we are killing it from watcher goroutine from symfony. My strong suspicion is these leads to inconsistent behavior because of terminal's foreground behavior and read / write / piping std{in,our,err} of parent process and child process.

[...]

PR #405, opened 2024-01-04
Set SysProcAttr Foreground: true when user run cmd. Fixes: #404

sarim:
Fixes: #404

Details description at linked issue. Setting  `SysProcAttr` `Setpgid: true` in a previous commit https://github.com/symfony-cli/symfony-cli/commit/ff1958fbb1f875372c8d654e37f4c7c29327e9f2 broke interactivty with `symfony run`. With things like `symfony run psql`.

This PR sets Foreground based on runmode. It works in linux. Should work in macos too hopefully.

Not exactly related to this issue/PR, but runner_windows.go is basically dummy. So not sure worker lifecycle/signals/docker integration etc.. works perfectly in windows.

### DIFF
--- a/local/runner.go
+++ b/local/runner.go
@@ -277,7 +277,7 @@ func (r *Runner) buildCmd() (*exec.Cmd, error) {
 	cmd.Env = os.Environ()
 	cmd.Dir = r.pidFile.Dir
 
-	if err := buildCmd(cmd); err != nil {
+	if err := buildCmd(cmd, r.mode == RunnerModeOnce); err != nil {
 		return nil, err
 	}
 
--- a/local/runner_posix.go
+++ b/local/runner_posix.go
@@ -8,11 +8,12 @@ import (
 	"syscall"
 )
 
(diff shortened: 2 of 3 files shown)


fabpot:
Thank you @sarim

tgalopin:
Hi @sarim ! I think maybe https://github.com/symfony-cli/symfony-cli/issues/409 is related to this, maybe we should also use the interactivity detection (flag/terminal) when populating the Foreground argument?

fabpot:
Reverted fornow (with #410 as well) as it doesn't work (at least on Macos where it hangs and ^C doesn't work anymore).

sarim:
> Reverted fornow (with #410 as well) as it doesn't work (at least on Macos where it hangs and ^C doesn't work anymore).
> 

@fabpot interesting, so "symfony run psql" works for you in macos without setting `Foreground`? Do postgresql prompt appears and you can write a query, press enter and get results back? If so then I can rework to set foreground on only linux.

PR #411, opened 2024-01-15
feat: add an option to create a new project using Symfony Docker

dunglas:
The Core Team privately discussed repurposing the `--docker` flag of the `symfony new` command to create a project using [Symfony Docker](https://github.com/dunglas/symfony-docker). The goals are to have an official way to bootstrap a project using Docker with ease and to set up [FrankenPHP](https://frankenphp.dev) with Symfony.

This patch only covers project creation.
In the feature, it could be nice also to detect if Symfony Docker is used and forward `symfony console`, `symfony composer`, and similar commands to the Docker container.

Usage:

```console
symfony new --docker
```

cc @weaverryan

### DIFF
--- a/commands/local_new.go
+++ b/commands/local_new.go
@@ -66,7 +66,7 @@ var localNewCmd = &console.Command{
 		&console.BoolFlag{Name: "webapp", Usage: "Add the webapp pack to get a fully configured web project"},
 		&console.BoolFlag{Name: "api", Usage: "Add the api pack to get a fully configured api project"},
 		&console.BoolFlag{Name: "book", Usage: "Clone the Symfony: The Fast Track book project"},
-		&console.BoolFlag{Name: "docker", Usage: "Enable Docker support"},
+		&console.BoolFlag{Name: "docker", Usage: "Use the Symfony Docker Skeleton (includes FrankenPHP)"},
 		&console.BoolFlag{Name: "no-git", Usage: "Do not initialize Git"},
 		&console.BoolFlag{Name: "upsun", Usage: "Initialize Upsun configuration"},
 		&console.BoolFlag{Name: "cloud", Usage: "Initialize Platform.sh configuration"},
@@ -110,6 +110,7 @@ var localNewCmd = &console.Command{
 		}
 
 		symfonyVersion := c.String("version")
+		symfonyDocker := c.Bool("docker")
 
 		if c.Bool("book") {
 			if symfonyVersion == "" {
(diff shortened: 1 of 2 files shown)


weaverryan:
I’m not qualified to review technically, but this is exactly what I had in mind. It will help with discoverability.

> it could be nice also to detect if Symfony Docker is used and forward symfony console, symfony composer, and similar commands to the Docker container.

That’s super interesting!

sarim:
please excuse my interjection as I'm not a core team member, but I'd like to share some thoughts on the proposed `--docker` flag. I appreciate the effort to integrate Docker seamlessly, but automatically using "dunglas/symfony-docker" and FrankenPHP could be confusing for those less familiar with Docker or web server administration. It would be ideal if the default behavior created a project with de-facto-standard technologies: nginx/php-fpm. Perhaps the `--docker` flag could require specifying a skeleton name or URL, initializing the project accordingly. Official Symfony skeletons should be created in official namespaces like `symfony/docker-fpm` or `symfony/docker-frankenphp`. Then these official skeletons would be listed in the `--help` output. Additionally, a Symfony documentation article outlining the required structure for Docker containers to be compatible with this Symfony-CLI feature would be beneficial. This could enable the community to create and utilize their own skeletons effectively.

I apologize again for intruding on a core team discussion.

weaverryan:
Hi!

Valid points! From a practical point of view, I’m not sure we can maintain 2 different docker skeletons. And so, the one will be opinionated.

 > Official Symfony skeletons should be created in official namespaces like symfony/docker-fpm or symfony/docker-frankenphp

Being under the symfony org makes sense to me. But @dunglas is the one doing the maintaining of the docker skeleton in reality. But yea, I think it’d be cool to have it under the symfony org if it’s the official option.

dunglas:
@sarim Unfortunately, FPM is not well suited to containerized platforms. This is one of the main reasons I started working on FrankenPHP: to easily distribute Symfony applications as Docker images, as easily as with Node or Go. As a "Symfony Docker" maintainer for years (who used NGINX and FPM in the early days), I can guarantee that using Caddy and then FrankenPHP has simplified a lot of things and greatly improved the developer's experience. I gave [a full presentation on this topic] (https://dunglas.dev/2022/10/frankenphp-the-modern-php-app-server-written-in-go/) at SymfonyCon 2022.

@weaverryan I'm not against moving Symfony Docker under the Symfony namespace, as long as we can continue to grant maintenance rights to @maxhelias, who has done a great job maintaining it for a very long time.

PR #413, opened 2024-01-18
Replace hardcoded temp directory path with the POSIX variable

malefici:
The [POSIX variable](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) `$TMPDIR` instead of `/tmp`.

I tried installing Symfony CLI on Termux, but I got an error from `curl` about `/tmp/symfony....` file. The problem is that the real path for the temporary directory is not `/tmp`.

### DIFF
--- a/installer/bash-installer
+++ b/installer/bash-installer
@@ -166,10 +166,10 @@ latest_url=${CLI_DOWNLOAD_URL_PATTERN/~platform~/${platform}}
 output "  Downloading ${latest_url}...";
 case $downloader in
     "curl")
-        curl --fail --location "${latest_url}" > "/tmp/${CLI_TMP_NAME}.tar.gz"
+        curl --fail --location "${latest_url}" > "${TMPDIR}/${CLI_TMP_NAME}.tar.gz"
         ;;
     "wget")
-        wget -q --show-progress "${latest_url}" -O "/tmp/${CLI_TMP_NAME}.tar.gz"
+        wget -q --show-progress "${latest_url}" -O "${TMPDIR}/${CLI_TMP_NAME}.tar.gz"
         ;;
 esac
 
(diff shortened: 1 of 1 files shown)


fabpot:
I would still fallback to `/tmp` if the env var does not exist.

sarim:
Maybe something like
```bash
SF_TMPDIR="${TMPDIR:-/tmp}"
```
Somewhere in the beginning of the file, then use $SF_TMPDIR variable in all later stages.

fabpot:
> Maybe something like
> 
> ```shell
> SF_TMPDIR="${TMPDIR:-/tmp}"
> ```
> 
> Somewhere in the beginning of the file, then use $SF_TMPDIR variable in all later stages.

Exactly

malefici:
Pls, check. I chose the variable name `CLI_TMPDIR`. @fabpot, why do you prefer to fallback to a hardcoded value? `$TMPDIR` is a POSIX variable and must exist in every fully or partly POSIX-compatible OS. Have you ever seen such a case?

fabpot:
> Pls, check. I chose the variable name `CLI_TMPDIR`. @fabpot, why do you prefer to fallback to a hardcoded value? `$TMPDIR` is a POSIX variable and must exist in every fully or partly POSIX-compatible OS. Have you ever seen such a case?

Just my experience. When I read "it's not possible", I think "A bug is going to be reported within 24 hours after the merge".

fabpot:
Thank you.

tucksaun:
> `$TMPDIR` is a POSIX variable and must exist in every fully or partly POSIX-compatible OS. Have you ever seen such a case?

Just a single example: run an Alpine docker container and it won't be available 😉

sarim:
> `$TMPDIR` is a POSIX variable and must exist in every fully or partly POSIX-compatible OS. Have you ever seen such a case?

I just checked few of my machines, Ubuntu and Fedora running on barebones, debian and ubuntu VMs running on ec2, linode, digitalocean, ubuntu and debian running in WSL2, `TMPDIR` wasn't set in any of those machines :P

RafaelKr:
@malefici Just recently I watched this YouTube video "How A Steam Bug Deleted Someone’s Entire PC": https://www.youtube.com/watch?v=qzZLvw2AdvM

TLDW: The steam installer ran the command `rm -rf $STEAMROOT/*` but due to a bug `$STEAMROOT` was undefined, so actually it ran `rm -rf $STEAMROOT/*` which deleted all files in the root directory including his 3TB backup drive which was mounted somewhere at `/mnt`.

Of course this won't happen in this case. But still it's always advisable to make sure a variable is initialized with a sane default value.

malefici:
@tucksaun @sarim @RafaelKr 
Thank you a lot for the examples. I will be much more careful with env vars.

> Ubuntu and Fedora running on barebones
o_O This var exists on my nodes.

PR #415, opened 2024-01-19
Replace unmaintained `github.com/hpcloud/tail` with `github.com/nxadm/tail`

principis:
Replaces the unmaintained `github.com/hpcloud/tail` with the active fork `github.com/nxadm/tail`.

This solves a Fedora packaging issue at the same time.


fabpot:
Thank you

Issue #425, opened 2024-02-05
Upgrading to 5.8.6 causes invalid MariaDB version string in local docker environment


Jontsa:
For development, we are using MariaDB 10.6 in docker container and Symfony CLI to serve pages. After upgrading to Symfony CLI 5.8.6, we started seeing deprecation warnings like this from Doctrine:

`MySQL 5.6 support is deprecated and will be removed in DBAL 4. Consider upgrading to MySQL 5.7 or later.`

It looks like 5.8.6 is unable to parse the version string from MariaDB server correctly. If I debug $_ENV["DATABASE_URL"], it shows `mysql://***:***@127.0.0.1:32769/***?sslmode=disable&charset=utf8mb4&serverVersion=1:10.6.15+maria~ubu2004` and 
AbstractMySQLDriver receives version string `1:10.6.15 maria~ubu2004`.

Downgrading to 5.8.4 fixes the issue.

## How to reproduce

docker-compose.yml
```yaml
version: '3.7'

services:
  database:
    container_name: database
    image: mariadb:10.6
...
```

.env
```env
###> doctrine/doctrine-bundle ###
DATABASE_URL=mysql://db_user:db_password@127.0.0.1:3306/sirukyc?serverVersion=10.6.11-MariaDB
###< doctrine/doctrine-bundle ###
```

config/packages/doctrine.yaml
```yaml
doctrine:
    dbal:
        default_connection:       default
        connections:
            default:
                profiling_collect_backtrace: '%kernel.debug%'
                url: '%env(resolve:DATABASE_URL)%'

        # IMPORTANT: You MUST configure your server version,
        # either here or in the DATABASE_URL env var (see .env file)
        #server_version: '16'
```

How MariaDB reports version string

```bash
$ docker compose exec database mysql -uroot -p
Enter password: 
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 106
Server version: 10.6.15-MariaDB-1:10.6.15+maria~ubu2004 mariadb.org binary distribution

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> 
```
```bash
$ docker compose exec database mysql --version
mysql  Ver 15.1 Distrib 10.6.15-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
```

Jontsa:
[...]

PR #432, opened 2024-02-09
Use doctrine.dbal.connections.default.server_version in addition to doctrine.dbal.server_version

fabpot:


nicolas-grekas:
DoctrineBundle already handles the merging of the yaml and the DSN so we should like remove all the yaml-related logic here:
https://github.com/doctrine/DoctrineBundle/blob/2.11.x/ConnectionFactory.php#L251

fabpot:
Fixed #425

fabpot:
> DoctrineBundle already handles the merging of the yaml and the DSN so we should like remove all the yaml-related logic here: https://github.com/doctrine/DoctrineBundle/blob/2.11.x/ConnectionFactory.php#L251

We are always specifying a version in the URL env var. So, the version set in `doctrine.yaml` can never be used.

PR #435, opened 2024-02-11
Add first integration tests

alexandre-daubois:
After a quick chat with @tucksaun, here's a try on adding a few integration tests. I only added a few ones to validate the way of doing this before going further. 🙂

### DIFF
--- a/integration/local_server_list_test.go
+++ b/integration/local_server_list_test.go
@@ -1,0 +1,65 @@
+//go:build integration
+// +build integration
+
+/*
+ * Copyright (c) 2024-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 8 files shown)


integration/local_server_list_test.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func TestServerList(t *testing.T) {
+	startCommands := []string{
+		"../symfony-cli server:start -d --dir=phpinfo/",
+		"../symfony-cli server:start -d --dir=hello_world/",
+	}
+
+	for _, command := range startCommands {
+		cmd := exec.Command(strings.Split(command, " ")[0], strings.Split(command, " ")[1:]...)
```
tucksaun, line 39: to get more control I would rather directly use the code here (starting the server in a goroutine and issuing http request, worst case: starting the application instead of directly the HTTP server).

tucksaun:
thank you for starting the work on this!

integration/local_server_list_test.go
alexandre-daubois: Would you do this in all functional tests or only this one?

PR #439, opened 2024-02-14
Fix MariaDB version when using Docker

### LINKED ISSUE #438: v5.8.8 - "symfony console make:migration" is broken
Framework version:  v6.4.3 with MariaDb running in Docker. I have the proper connection settings made (server version & charset)

Running `symfony console make:migration` fails to work properly and the error reported is 

~~~text
The metadata storage is not up to date, please run the sync-metadata-storage command to fix this issue. 
~~~

Running `./bin/console make:migration`  works as expected. Downgrading to v5.8.7 works also.

fabpot:
Closes #438

### DIFF
--- a/envs/envs.go
+++ b/envs/envs.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


envs/envs.go
```diff
@@ -186,8 +187,9 @@ func extractRelationshipsEnvs(env Environment) Envs {
 						// Docker support
 						if v, ok := endpoint["version"]; ok {
 							values[versionKey] = v.(string)
-							if strings.HasPrefix(values[versionKey], "1:") {
-								values[versionKey] = values[versionKey][2:]
+
+							if matches := regexp.MustCompile("^1\\:(\\d+\\.\\d+\\.\\d+)").FindStringSubmatch(values[versionKey]); matches != nil {
```
nicolas-grekas, line 191:
```suggestion
							if matches := regexp.MustCompile("^\d:(\\d+\\.\\d+\\.\\d+).maria").FindStringSubmatch(values[versionKey]); matches != nil {
```

Issue #451, opened 2024-03-21
CLI v5.8.12 (still) does not create the proper env var for MongoDB!


malhusseiny:
A MongoDB service with the following docker-compose configuration in `compose.yml`:

```yaml
mongo:
  image: mongo
  restart: always
  environment:
    MONGO_INITDB_ROOT_USERNAME: ${MONGO_USER:-app}
    MONGO_INITDB_ROOT_PASSWORD: ${MONGO_PASSWORD:-!ChangeMe!}
    MONGO_INITDB_DATABASE: ${MONGODB_DB:-symfony}
```

causes the variables `MONGO_DATABASE`, `MONGO_DB` , `MONGO_NAME` to be set to `symfony`, which is fine. However, it also sets the `authSource` param of `MONGO_URL` to `symfony` as shown below:

```shell
symfony var:export --multiline             
export MONGO_DATABASE=symfony
export MONGO_DB=symfony
export MONGO_HOST=127.0.0.1
export MONGO_NAME=symfony
export MONGO_PASSWORD=!ChangeMe!
export MONGO_PORT=32795
export MONGO_SCHEME=mongodb
export MONGO_SERVER=mongodb://127.0.0.1:32795
export MONGO_URL=mongodb://app:!ChangeMe!@127.0.0.1:32795/?authSource=symfony
export MONGO_USER=app
export MONGO_USERNAME=app
```

Without the `MONGO_INITDB_DATABASE: ${MONGODB_DB:-symfony}` in the docker compose file, all database-related environment variables as well as the `authSource` param become empty. It would be great if you could untangle that coupling.

fabpot:
The logic that parses the Docker configuration for MongoDB is here: https://github.com/symfony-cli/symfony-cli/blob/main/envs/docker.go#L401 and the logic to use that information to create env vars is here: https://github.com/symfony-cli/symfony-cli/blob/main/envs/docker.go#L401

As I'm not a MongoDB user myself, I'm not sure what should be changed there. @malhusseiny Can you help me?

fabpot:
But here, you've completely removed the "path", which looks wrong to me. But then again, I'm not a MongoDB user, so ideally, I'd like others to chime in here.
As a first step, maybe you can create a PR for that change?

PR #458, opened 2024-03-27
fix: Remove incorrect `authSource` param from `MONGODB_URL` environment variable

malhusseiny:
This fix aims to address the issue with the incorrect MongoDB connection string specified in the `MONGODB_URL` environment variable, as generated by the CLI. #451 

Consider a Docker Compose configuration for a MongoDB service as follows:

```yaml
mongodb:
  image: mongo:latest
  environment:
    MONGO_INITDB_ROOT_USERNAME: app
    MONGO_INITDB_ROOT_PASSWORD: changeMe
    MONGO_INITDB_DATABASE: symfony
```

This setup results in the `symfony var:export` command generating a `MONGODB_URL` environment variable similar to:

```sh
export MONGODB_URL=mongodb://admin:changeMe@127.0.0.1:27017/?authSource=symfony
```

Such a configuration typically leads to authentication failures for most applications attempting to connect to MongoDB.

To address this, I have chosen to omit the "path" portion of the connection string entirely, rather than merely replacing `/authSource=symfony` with `/symfony`. The latter approach would not resolve the underlying issue. In MongoDB connection strings, specifying a default working database in the URL path (e.g., `mongodb://user:pass@127.0.0.1:27017/defaultDb`) uses `defaultDb` as both the default and authentication database. This behavior is often unintended. To correctly designate separate databases for default operations and authentication, the `authSource` parameter must be explicitly added (e.g., `mongodb://user:pass@127.0.0.1:27017/defaultDb?authSource=authDb`). However, this necessitates the declaration of an `authSource` database within the Docker Compose configuration. I propose avoiding this complication entirely, as both the default database and the `authSource` can be separately specified through dedicated options (e.g., in Doctrine) without embedding them in the connection string.

### DIFF
--- a/envs/envs.go
+++ b/envs/envs.go
@@ -239,7 +239,7 @@ func extractRelationshipsEnvs(env Environment) Envs {
 				if !isMaster(endpoint) {
 					continue
 				}
-				values[fmt.Sprintf("%sURL", prefix)] = fmt.Sprintf("%s://%s:%s@%s:%s/?authSource=%s", endpoint["scheme"].(string), endpoint["username"].(string), endpoint["password"].(string), endpoint["host"].(string), formatInt(endpoint["port"]), endpoint["path"].(string))
+				values[fmt.Sprintf("%sURL", prefix)] = fmt.Sprintf("%s://%s:%s@%s:%s", endpoint["scheme"].(string), endpoint["username"].(string), endpoint["password"].(string), endpoint["host"].(string), formatInt(endpoint["port"]))
 				values[fmt.Sprintf("%sSERVER", prefix)] = formatServer(endpoint)
 				values[fmt.Sprintf("%sHOST", prefix)] = endpoint["host"].(string)
 				values[fmt.Sprintf("%sPORT", prefix)] = formatInt(endpoint["port"])


tucksaun:
I see why not defining the authSource might make sense.
Yet, the official Docker image seems to hardcode the authentication database to be `admin` (according to their Docker Hub doc and the [entrypoint](https://github.com/docker-library/mongo/blob/f8a378ac3249bb74529b88ad084f6c547f2b4b19/7.0/docker-entrypoint.sh#L372)) so maybe we could define it when we detected some credentials to make it a smooth experience for most users?

Regarding the main database I'm not sure to understand why we should not define it. Because if I get that correctly this means this will make any user of mongodb required to explicitly specify it in Doctrine's configuration, right?

tucksaun:
I think it might also be interesting to ping @quentint who initially contributed the addition of the `MONGODB_URL` definition :)

jmikola:
:wave: @alcaeus is on holiday so @gromnan summoned me instead.

> In MongoDB connection strings, specifying a default working database in the URL path (e.g., `mongodb://user:pass@127.0.0.1:27017/defaultDb`) uses defaultDb as both the default and authentication database.

With respect to the PHP driver, there is no notion of a "default database" and it'd be helpful to only consider an "auth database" (with preference given to the `authSource` URI option over the "path" component in the URI. The [MongoDB Connection String spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.md#auth-database-optional) is the authority here.

That said, some _other_ drivers/libraries _may_ still infer a default database from the "path" component in the URI (see https://github.com/mongodb/mongo-php-driver/issues/1470#issuecomment-1721299444 for more context on that); however, this isn't relevant to Doctrine ODM.

According to [`DoctrineMongoDBExtension::loadConnections()`](https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.0.1/src/DependencyInjection/DoctrineMongoDBExtension.php#L363), the `server` property for a connection is only used to construct the PHP library's [`MongoDB\Client`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBClient__construct/) object. The connection string seems entirely unrelated to any default database utilized for Doctrine ODM models.

In the absence of specifying a database name in the model metadata, ODM relies on `Configuration::getDefaultDB()`, and falls back to `'doctrine'` (see: [`DocumentManager::getDocumentDatabase()`]](https://github.com/doctrine/mongodb-odm/blob/2.7.0/lib/Doctrine/ODM/MongoDB/DocumentManager.php#L319)). When using the Symfony bundle, `Configuration::setDefaultDB()` should always be called with _something_. If the application doesn't explicitly define `default_database` in the bundle config, the DIC configuration uses `'default'` (see: [`Configuration::getConfigTreeBuilder()`](https://github.com/doctrine/DoctrineMongoDBBundle/blob/5.0.1/src/DependencyInjection/Configuration.php#L81C17-L81C81)).

> Yet, the official Docker image seems to hardcode the authentication database to be admin (according to their Docker Hub doc and the entrypoint) so maybe we could define it when we detected some credentials to make it a smooth experience for most users?

I see no reason to specify `'admin'` as a default value, as drivers already use that as a default for database-related auth mechanisms (see: [MongoDB Auth spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#auth-related-options). Note that other auth mechanisms might use a different value (e.g. `$external`), but I expect those wouldn't come up with the Docker container being used here.

----

Some other thoughts:

 * I don't understand what the `isMaster()` function (called just above the diff in this PR) does. The function dates back to the first commit for `envs.go` in https://github.com/symfony-cli/symfony-cli/commit/5ac4bd9fb0721e18bb478fb2fd86831d483a47bd#diff-8a87b46fa3ed09eda7bafda20abcc8a2dde06d513c4a38a1a1ef8376b24adb14R307 and originally applied to pgsql, mysql, and mongodb schemes. It was later removed for pgsql and mysql in https://github.com/symfony-cli/symfony-cli/commit/1c0b3c8c3fb6595a5e87d3176377f9add543d023#diff-8a87b46fa3ed09eda7bafda20abcc8a2dde06d513c4a38a1a1ef8376b24adb14L125 but remains in place for MongoDB.

 * I also don't understand what the other environment variables are used for. `MONGO_NAME`, `MONGO_DATABASE`, and `MONGO_DB` are all still set to `endpoint["path"]`. Since we previously established that the "path" URI component is _not_ a default database but merely a deprecated way of specifying the auth databsae, I wouldn't expect it to be used for any other purpose. And even if it was desirable to expose the "path" as an environment variable, assigning it to three different vars seems redundant and more likely to mislead users. This question relates more to the paste that @malhusseiny included in https://github.com/symfony-cli/symfony-cli/issues/451.

fabpot:
Fixed by #649

PR #462, opened 2024-04-06
Add `--no-workers` option

### LINKED ISSUE #459: Allow to server:start without workers
Maybe add a `--no-workers` option?

This is an alternative solution to #399 for CI environments, where building may already be done, watching files may not be needed, and consuming message be done in another way like by using CI services.

yguedidi:
closes #459


fabpot (approved):

fabpot:
@yguedidi Can you maybe add a note in the docs as well about this new flag?

PR #463, opened 2024-04-06
Ignore generated binary

yguedidi:
I committed this file by mistake in my first PR, so maybe it's worth ignoring it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /dist
 /vendor
+symfony-cli


tucksaun (approved):

fabpot:
I don't think this is useful as this name is not referenced anywhere, so when you compile the binary yourself, you can use any name.

yguedidi:
> I don't think this is useful as this name is not referenced anywhere, so when you compile the binary yourself, you can use any name.

@fabpot I just run `go generate` then `go build` and this the filename I got

fabpot:
> > I don't think this is useful as this name is not referenced anywhere, so when you compile the binary yourself, you can use any name.
> 
> @fabpot I just run `go generate` then `go build` and this the filename I got

Of course, I never run `go build` without `-o` :)

yguedidi:
> > > I don't think this is useful as this name is not referenced anywhere, so when you compile the binary yourself, you can use any name.
> > 
> > 
> > @fabpot I just run `go generate` then `go build` and this the filename I got
> 
> Of course, I never run `go build` without `-o` :)

Because you're way more experienced in Go, me it was the first time writing Go, not even in tutorials, I'm quite proud even for just few lines :)

PR #475, opened 2024-05-08
Hanging fix

fabpot:
Reverting the revert

### DIFF
--- a/local/runner.go
+++ b/local/runner.go
@@ -236,8 +236,6 @@ func (r *Runner) Run() error {
 		// Command exited
 		case err := <-cmdExitChan:
 			logger.Debug().Msg("Received exit")
-			err = errors.Wrapf(err, `command "%s" failed`, r.pidFile)
-
 			if err == nil && r.SuccessHook != nil {
 				logger.Debug().Msg("Running success hook")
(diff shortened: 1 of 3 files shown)


fabpot:
Tested on Linux and MacOS and it seems to behave correctly.

fabpot:
Indeed, I've reverted most of #308 by @tucksaun.

fabpot:
I've again tested this patch extensively and I don't see any issues. I might miss something, but I think this is pretty solid now. At least, much better than the current state.

local/runner.go
```diff
@@ -254,36 +252,23 @@ func (r *Runner) Run() error {
...
-			case sig := <-sigChan:
-				logger.Info().Msgf(`Signal "%s" received, exiting`, sig)
-				return nil
-			case <-restartChan:
-				logger.Debug().Msg("Received restart")
-				timer.Stop()
-			case <-timer.C:
-				logger.Debug().Msg("Received timer message")
-			}
+			return nil
```
tucksaun, line 271: I have the feeling that returning nil here means a command not setup to always restart (hence everything else than PHP SAPIs) will not restart when done. Am I correct?

Issue #117, opened 2022-03-22
A problem with "firewall incoming requests" in macOS


javiereguiluz:
In the previous closed-source version of Symfony CLI, the first time I run it, I saw this dialog to allow/deny incoming requests in the macOS firewall:

![firewall-popup](https://user-images.githubusercontent.com/73419/159443916-08a87b66-3212-4889-b958-1e808ac6f9ad.jpeg)

After allowing it, I added the binary to the list of applications which are allowed incoming requests and that's it. I never saw that firewall dialog again.

-----

With the new open source version, I see the same firewall dialog whenever I run `symfony proxy:start` and `symfony serve -d`. I've tried adding the Symfony CLI binary to the list of allowed apps (both the real binary at e.g. `/opt/homebrew/Cellar/symfony-cli/5.4.2/bin/symfony` and the link to the binary at e.g. `/opt/homebrew/bin/symfony`) and I update this whenever I upgrade the Symfony CLI version, but it doesn't work.

Today I was really tired of this issue and I looked for other solutions. I found this explanation for the problem --> https://superuser.com/a/912663 and this solution for it --> https://superuser.com/a/1438094

Running the following command solved the problem for me:

```
sudo codesign --force --deep --sign - /opt/homebrew/Cellar/symfony-cli/5.4.2/bin/symfony
```

Just sharing this here in case others have the same problem or in case we can do something in this project to improve the situation. Thanks!

Kocal:
Thanks @javiereguiluz, some of my team members and me faced the same issue!

gndk:
Same problem here on macOs 12.2.1 and Symfony CLI 5.4.2. I think this might be a macOs Problem, not Symfony CLI, because I have the same problem when running Node stuff, for example on every `encore dev`.

Your fix worked for me, too. Thanks! At least for Symfony CLI. Can't get it to work on Node for now.

fabpot:
@javiereguiluz Is it something we can add to the Symfony docs somewhere (and also maybe on the download page)?

tucksaun:
maybe the Homebrew Formula can be tweaked to do it automatically on install?

javiereguiluz:
Closing as fixed in the docs: https://github.com/symfony/symfony-docs/pull/17881

tucksaun:
Thank you @javiereguiluz for documenting this

PR #494, opened 2024-07-06
Only listen on 127.0.0.1 not all addresses

### LINKED ISSUE #319: Bind Address causes popup on mac
It would be great if either the bind address could be specified via a option or it just defaults to 127.0.0.1

Currently symfony serve seems to default to 0.0.0.0 which means external connections are allowed and it will cause this annoying popup on macOS 13.3.1 each time it's started. 

From a security standpoint it makes more sense anyway to default to 127.0.0.1
The usecase of external connections is very slim, IMO. 

image

gempir:
This is a security risk, because you are exposing your application to other devices on your network if your device has no firewall configured. 

More importantly though this fix allows root free running of symfony and on macOS you will no longer get an annoying popup to allow incoming connections.

Fixes #319

This already was part of confusion and discussion in #117

### DIFF
--- a/local/process/listener.go
+++ b/local/process/listener.go
@@ -39,15 +39,9 @@ func CreateListener(port, preferredPort int) (net.Listener, int, error) {
 		max = 1
 	}
 	for {
-		// we really want to test availability on 127.0.0.1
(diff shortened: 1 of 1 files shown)


gempir:
Can I do anything further to help this getting merged?

PR #500, opened 2024-07-17
Add support for MySQL/MariaDB/PostgreSQL HA

fabpot:


envs/envs.go
```diff
@@ -123,7 +123,8 @@ func extractRelationshipsEnvs(env Environment) Envs {
 			}
 			prefix = strings.Replace(prefix, "-", "_", -1)
 
-			if scheme == "pgsql" || scheme == "mysql" {
+			// we support mysql and mysql-replica
+			if scheme == "pgsql" || strings.HasPrefix(scheme.(string), "mysql") {
```
tucksaun, line 127: what do you think of moving the type check to `scheme` declaration: `scheme, _ := endpoint["scheme"].(string)` ?

fabpot: I tried :) But it means many changes that are not worth it (especially as we are reading the scheme from endpoint in some other functions that would not get the changes we're doing for postgresql for instance).

Theosakamg:
+1

PR #505, opened 2024-07-19
Return scheme for undefined docker container

kevin-studer:
This PR should fix #504 

In the issue I talked about to setting the scheme to "tcp". While it would work, it is not entirely accurate. So I've returned the scheme `undefined` (as a string), which matches closer to what it is actually happening.

PR #507, opened 2024-07-19
Implement test which checks env relationships

kevin-studer:
Hi there

In #505 I've talked about testing the docker integration of the symfony cli more sophisticated. I've implement a proposal which includes [testcontainers](https://golang.testcontainers.org/) which start specific containers and checks if the symfony cli generates the correct variables.

The downside of this approach is that the developer which runs tests should have a docker host available (either local installed or configured with the env variable `DOCKER_HOST`).

Also I'm quit unsure if the tests should be refactored. This is due the fact that I haven't a lot of experience in Go, so perhaps someone can look into it and guide me in the right direction, if the tests are ok.

Thanks!

### DIFF
--- a/commands/resources/completion.bash
+++ b/commands/resources/completion.bash
@@ -46,7 +46,7 @@ _complete_{{ .App.HelpName }}() {
     done
 
     # Use newline as only separator to allow space in completion values
-    IFS=$'\n'
+    local IFS=$'\n'
 
     local cur prev words cword
     _get_comp_words_by_ref -n := cur prev words cword


tucksaun:
Sounds like a good idea 👍

envs/local_test.go
```diff
@@ -98,3 +107,261 @@ func (s *LocalSuite) TestRelationships(c *C) {
 		"PGHOST":                 "127.0.0.1",
 	})
 }
+
+func (s *LocalSuite) TestGenericContainer(c *C) {
+	ctx := context.Background()
+
+	composeProject := os.Getenv("COMPOSE_PROJECT_NAME")
```
tucksaun, line 114: looks to me that it would be more realistic (and easier to debug?) to use their Docker Compose module: https://golang.testcontainers.org/features/docker_compose/

PR #516, opened 2024-09-10
Implement test which checks env relationships

kevin-studer:
Hi there

This PR continues the work from https://github.com/symfony-cli/symfony-cli/pull/507. I've made a mistake and closed the PR by accident.

> In https://github.com/symfony-cli/symfony-cli/pull/505 I've talked about testing the docker integration of the symfony cli more sophisticated. I've implement a proposal which includes [testcontainers](https://golang.testcontainers.org/) which start specific containers and checks if the symfony cli generates the correct variables.
> The downside of this approach is that the developer which runs tests should have a docker host available (either local installed or configured with the env variable DOCKER_HOST).
> Also I'm quit unsure if the tests should be refactored. This is due the fact that I haven't a lot of experience in Go, so perhaps someone can look into it and guide me in the right direction, if the tests are ok.
> Thanks!

I've reworked the testcontainers integration by providing `docker-compose` files for each scenario. It should help maintaining and debugging this test.

Thanks for looking into it. Feedback is very welcome!


kevin-studer:
@tucksaun I've implemented docker-compose module as requested. Sorry for the inconvenience of the 2 PRs.

PR #525, opened 2024-10-11
Allow to define the IP on which `server:start` should listen to

### LINKED ISSUE #523: PR #494 broke Docker startup with `ENTRYPOINT ["symfony", "server:start", "--no-tls", "--port=80"]`
The recent pull request #494, which changes the default behavior to only listen on 127.0.0.1, has caused issues when running Symfony CLI in a Docker container. Specifically, the following `ENTRYPOINT` command no longer works as expected:

```Dockerfile
ENTRYPOINT ["symfony", "server:start", "--no-tls", "--port=80"]
```

### Steps to Reproduce:  
Create a Dockerfile with the following content:
```
FROM php:8.2-fpm-alpine3.17
RUN   curl -s https://get.symfony.com/cli/installer | bash \
	&& mv /root/.symfony5/bin/symfony /usr/local/bin/symfony
WORKDIR /var/www/html
ENTRYPOINT ["symfony", "server:start", "--no-tls", "--port=80"]
```
Build and run the Docker container.
Attempt to access the Symfony server from outside the container.
### Expected Behavior: 
The Symfony server should be accessible from outside the Docker container.  Actual Behavior: The server is not accessible because it only listens on 127.0.0.1.  
### Proposed Solution:
 Allow specifying the address to listen on via an environment variable or command-line option, or revert the default behavior to listen on all addresses (0.0.0.0).  

### Additional Context: 
This issue affects users who rely on Docker for their development and deployment workflows. The change in default behavior has introduced a breaking change for these users.

### LINKED ISSUE #524: Add flag for all interfaces or specific interface ip
With the Merge request 494 the cli only allows connection via 127.0.0.1 but i need to share the website internaly with a non programmer on a different system without local dns resolution or proxy setup

tucksaun:
This PR introduces two additional flags to the `server:start` command: `listen-ip` allows to define precisely the IP one wants to use and defaults to “127.0.0.1”, and `allow-all-ip` allows to enable listening on all interfaces (the behavior before #494 gets merged).
 
In addition to the flags, one can also use the SYMFONY_LISTEN_IP and SYMFONY_ALL_IP environment variables.

I didn’t go the “listen address” path because it would break the port flag and the preferred port logic but can move towards this later one with some BC layer.

Refs #494
Fix #523, fix #524

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,8 @@ COPY symfony /usr/local/bin/
 
 FROM scratch
 
+ENV SYMFONY_ALLOW_ALL_IP=true
+
 ENTRYPOINT ["/usr/local/bin/symfony"]
 
 COPY --from=build . .
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -65,6 +65,8 @@ var localServerStartCmd = &console.Command{
 		&console.StringFlag{Name: "document-root", Usage: "Project document root (auto-configured by default)"},
 		&console.StringFlag{Name: "passthru", Usage: "Project passthru index (auto-configured by default)"},
 		&console.IntFlag{Name: "port", DefaultValue: 8000, Usage: "Preferred HTTP port"},
+		&console.StringFlag{Name: "listen-ip", DefaultValue: "127.0.0.1", Usage: "The IP on which the CLI should listen"},
+		&console.BoolFlag{Name: "allow-all-ip", Usage: "Listen on all the available interfaces"},
 		&console.BoolFlag{Name: "daemon", Aliases: []string{"d"}, Usage: "Run the server in the background"},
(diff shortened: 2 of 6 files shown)


fabpot:
Looks good to me, thank you @tucksaun

scadergit:
Working great, thanks for the quick update.

IndraGunawan:
hi @tucksaun , thanks for this PR.

i have a question, is there a way to set this config as default? so I don't need to add `--allow-all-ip` every time I do `symfony serve -d`

tucksaun:
> hi @tucksaun , thanks for this PR.
> 
> 
> 
> i have a question, is there a way to set this config as default? so I don't need to add `--allow-all-ip` every time I do `symfony serve -d`

No config files to do this at the moment. But you can add `SYMFONY_ALLOW_ALL_IP=trur` to your environment to achieve this (see the change in the Dockerfile that does it)

PR #530, opened 2024-10-26
feat: consider php.ini from Nix

shyim:
When you create a php.ini in local dir and use PHP from nix. All extensions are gone as Nix sets a PHP_INI_SCAN_DIR

### DIFF
--- a/local/php/executor.go
+++ b/local/php/executor.go
@@ -272,6 +272,17 @@ func (e *Executor) Config(loadDotEnv bool) error {
 		if e.iniDir != "" {
 			dirs += string(os.PathListSeparator) + e.iniDir
 		}
+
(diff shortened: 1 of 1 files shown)


tucksaun:
shouldn’t this work out of the box as we preprend the directory to any `PHP_INI_SCAN_DIR` already set?

tucksaun:
If this is in the Nix wrapper, it is after we set `PHP_INI_SCAN_DIR`.
We don't override, instead, we prepend our value to preserve the original value of `PHP_INI_SCAN_DIR`.

It looks like Nix had something similar but decided to change their behavior in https://github.com/NixOS/nixpkgs/pull/221845.
I don't think it is our responsibility to ship a really specific Nix workaround based on their controversial decision.

fabpot:
Closing in favor of #719

PR #531, opened 2024-10-26
fix: let system level set env win

shyim:
when I have a .env

```
foo=1
```

and then do:
```
export foo=my-value

symfony php -i | grep foo
```

I see it is set to 1. and this is wrong. When I export a env variable it should win


tucksaun:
I’m wondering if instead we should not change the logic in `Executor.Execute()` to force the current environment to be appended instead of being preprended.
/cc @nicolas-grekas wdyt?

nicolas-grekas:
No opinion about the implementation, but the description of the issue looks correct to me, the precedence needs a fix.

PR #541, opened 2024-11-28
Fix path on local envs

fabpot:


nathanheffley:
This has caused a massive breaking change for any Symfony project that isn't a git repository itself. Is there a chance we could revisit it or improve the detection like in [#551](https://github.com/symfony-cli/symfony-cli/pull/551)?

fabpot:
See #555

PR #546, opened 2024-12-09
fix: current project directory and console path determinations should traverse upward

tucksaun:


fabpot:
Tests fail apparently.

jankal:
@tucksaun could you describe the issue you are fixing a little better? - maybe this intersects with #551

fabpot:
See #555

PR #551, opened 2024-12-17
feat(local): guess project dir based on presence of .symfony.local.yaml

jankal:
Recently `symfony-cli` does not discover my project roots anymore - this issue is most obvious in monorepos with different projects. Thus I thought about a different way to discover the project dir.

As `symfony-cli` can read config from `.symfony.local.yaml`, we should assume the directory containing this config file to be the project directory.

### DIFF
--- a/envs/local.go
+++ b/envs/local.go
@@ -286,8 +286,9 @@ func (l *Local) webServer() Envs {
 
 func guessProjectDir(dir string) string {
 	for {
-		f, err := os.Stat(filepath.Join(dir, ".git"))
-		if err == nil && f.IsDir() {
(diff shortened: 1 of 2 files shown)


nathanheffley:
I just ran into an issue where I was testing out a symfony project (specifically the Docker integration with doctrine) and it wasn't working. Turned out it was because I hadn't initiated a .git repo.

I am submitting a PR so that debug mode at the very least tells you it couldn't detect a root directory and to initialize a git repository, but that feels clunky. And what if someone uses Mercurial?

I like the idea of detecting a file that is actually Symfony related makes a lot more sense.

fabpot:
See #555 for an alternative approach.

PR #565, opened 2025-02-07
feat: automatically generate latest PHP version

tucksaun:
@fabpot do you want me to add an automation (as for the PSH commands generation)?
or do you prefer doing it by hand before tagging?


fabpot:
Let's automate.

tucksaun:
@fabpot automation added, you can see it running here: https://github.com/tucksaun/symfonycli/actions/runs/13385042472/job/37380106307

.github/workflows/go_generate_update.yml
```diff
@@ -5,6 +5,12 @@ on:
     schedule:
         - cron:  '0 8-18/4 * * 1-5'
 
+env:
+    GIT_AUTHOR_NAME: Tugdual Saunier
+    GIT_COMMITTER_NAME: Tugdual Saunier
+    GIT_AUTHOR_EMAIL: tugdual.saunier@gmail.com
+    GIT_COMMITTER_EMAIL: tugdual.saunier@gmail.com
```
fabpot, line 12: I like the idea of having a clearly identified bot user.

.github/workflows/go_generate_update.yml
```diff
@@ -19,29 +25,48 @@ jobs:
                 with:
                     go-version-file: 'go.mod'
             -
-                name: Prepare
+                name: Prepare, generate and format code
```
fabpot, line 28:
```suggestion
                name: Prepare, generate, and format code
```

.github/workflows/go_generate_update.yml
```diff
@@ -19,29 +25,48 @@ jobs:
...
                 run: |
                     rm -rf ~/.platformsh/bin/
-                    go generate ./
+                    go generate ./...
+                    go fmt ./...
```
fabpot, line 32: We need to exclude generated Go files (like `local/platformsh/commands.go`, `local/platformsh/config.go`)

.github/workflows/go_generate_update.yml
```diff
@@ -19,29 +25,48 @@ jobs:
...
             -
-                name: Commit and push the update
+                name: Commit the update for PSH-related code
                 if: steps.git.outputs.gitstatus != ''
                 run: |
-                    git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-                    git config user.name "github-actions[bot]"
                     git add local/platformsh/commands.go local/platformsh/config.go
-                    git commit -m "chore: Update supported Platform.sh services"
+                    git diff --staged --quiet || git commit -m "chore: Update supported Platform.sh services"
```
fabpot, line 55:
```suggestion
                    git diff --staged --quiet || git commit -m "Update supported Platform.sh services"
```

.github/workflows/go_generate_update.yml
```diff
@@ -19,29 +25,48 @@ jobs:
...
+
+            -   name: Commit the update for Docker-related code
+                if: steps.git.outputs.gitstatus != ''
+                run: |
+                    git add envs/docker_version.go
+                    git diff --staged --quiet || git commit -m "chore: Update Docker Client version"
```
fabpot, line 61:
```suggestion
                    git diff --staged --quiet || git commit -m "Update Docker Client version"
```

.github/workflows/go_generate_update.yml
```diff
@@ -19,29 +25,48 @@ jobs:
...
+
+            -   name: Commit the update for PHP-related code
+                if: steps.git.outputs.gitstatus != ''
+                run: |
+                    git add commands/php_version.go
+                    git diff --staged --quiet || git commit -m "chore: Update latest available PHP version"
```
fabpot, line 67:
```suggestion
                    git diff --staged --quiet || git commit -m "Update latest available PHP version"
```

.github/workflows/go_generate_update.yml
tucksaun: actually I would say the opposite but those files are annoyingly not properly formatted so each time un run `go fmt` I end up with local changes to discard :/

tucksaun: but I just though: or we could just fix the generator to have the generated file in a "clean" state.
I'm going to open another PR for this

.github/workflows/go_generate_update.yml
tucksaun: done

.github/workflows/go_generate_update.yml
tucksaun: updates

.github/workflows/go_generate_update.yml
tucksaun: applied

.github/workflows/go_generate_update.yml
tucksaun: see #576

.github/workflows/go_generate_update.yml
tucksaun: updated with the same account as before

Issue #397, opened 2023-12-18
Version 5.7.6 local:server:stop restarts fast-cgi instead of stopping local server


srsbiz:
![image](https://github.com/symfony-cli/symfony-cli/assets/1525842/1b9133cc-1544-4cc7-9903-33cec544dd8c)

![image](https://github.com/symfony-cli/symfony-cli/assets/1525842/8df35224-ccac-4e75-bf34-2104e02ec9af)

Rikijs:
Looks like i am seeing this on `Windows 10 Pro` with CLI `version 5.7.7`.
For me it started happening since CLI `version 5.7.6`.

![sf-error](https://github.com/symfony-cli/symfony-cli/assets/18629166/dc011d0a-df5f-45eb-9b42-5507c32a3a11)

srsbiz:
I can confirm that problem persists in 5.7.7 

I think this change is responsible: https://github.com/symfony-cli/symfony-cli/commit/ff1958fbb1f875372c8d654e37f4c7c29327e9f2#diff-2ae95fe2972282562f7c2b0ce5e64f8886c46c32664b5b939ba6133f9e6e7467R56-R63

Rikijs:
Still getting this error on `symfony-cli: 5.8.1` using PHP CGI `8.1.17`

srsbiz:
My hunch about change that caused this seems to be right, from stubs of go/src/os/exec.go

![Image](https://github.com/user-attachments/assets/8c5af3fe-5917-4f17-94e3-925de521a91b)

Can we have separate mechanism for windows to prevent restarting any workers? Like semaphore file maybe?

PR #566, opened 2025-02-08
fix: cleanup temporary directories

### LINKED ISSUE #383: Local server clutters .symfony5/tmp on windows
When starting local development server, new directory are created in `%USERPROFILE%/.symfony5/tmp` and few binaries are copied there. After terminating server with `CTRL+C` or by `symfony local:server:stop` this directory is not deleted. 

After 2+ years of usage I got ~28k directories with over 10GB of unused binaries

### LINKED ISSUE #553: $HOME/.symfony5/tmp is 18 GB, contains 4.272.171 files (folders?) and it's depleting the ext4 filesystem inodes (Linux)
I'm running **Symfony CLI version 5.10.4** (2024-10-11T14:05:11Z - stable) on a pre-production env. I use it to run custom **Symfony 6.3.12** commands and for composer.

The server runs **Ubuntu 22.04.5 LTS**, the filesystem is **ext4**  and the site is on NGINX and PHP-FPM, so the issue is not related to #383.

Today I got a "low inodes" alert, and it is due to the folder `$HOME/.symfony5/tmp`: it contains 4.272.171 files, for a total of 18 GB (the huge number of files is more important than the total size). `ls -la` takes a lot of time to compute, but then report something like this:

````
....
drwxr-xr-x 2 <username> www-data 4.0K Jun 24  2024 cpsdnr46fqoqma0f9c90/
drwxr-xr-x 2 <username> www-data 4.0K Jun 24  2024 cpsdntk6fqoqlsva0240/
drwxr-xr-x 2 <username> www-data 4.0K Jun 24  2024 cpsdo046fqoqkia7fv8g/
...
````
This is just a part of the list, and those folders were created `Jun 24  2024` (today is `29 Dec 2024`, so those folders are 6 months old).

I tried to ls a few folders, but they all look empty:

```
$ ls -la cpsdnr46fqoqma0f9c90
total 170448
drwxr-xr-x 2 <username> www-data      4096 Jun 24  2024 .
drwxr-xr-x 1 <username> www-data 174489600 Dec 29 09:51 ..
```

I run about a dozen of Symfony projects, all on different versions (from 4.4 to 7.2) and this is the first time I get this issue.

I'm glad to provide additional info.

### PR DESCRIPTION
**Pull request by tucksaun**
Fixes #383 
Fixes #553 

It also seems logical to work on cleaning any temporary directory while working on cleaning the `~/.symfony5/tmp` directory so this is a slightly bigger PR.
The second commit can extracted though if we want to focus on a single directory for now.

```bash
$ find ~/.symfony5/{tmp,var,log} -type d -mindepth 1 -maxdepth 1  | wc -l
       0
$ ~/Work/src/github.com/symfony-cli/symfony-cli/symfony-cli server:start

 [WARNING] The local web server is optimized for local development and MUST never be used in a production setup.



 [WARNING] Please note that the Symfony CLI only listens on 127.0.0.1 by default since version 5.10.3.
           You can use the --allow-all-ip or --listen-ip flags to change this behavior.



 [OK] Web server listening
      The Web server is using PHP FPM 8.4.3
      https://127.0.0.1:8000


Stream the logs via symfony-cli server:log
$ find ~/.symfony5/{tmp,var,log} -type d -mindepth 1 -maxdepth 1  | wc -l
       3
$ ~/Work/src/github.com/symfony-cli/symfony-cli/symfony-cli server:stop
Stopping PHP-FPM
Stopping Web Server


 [OK] Stopped 2 process(es) successfully


$ find ~/.symfony5/{tmp,var,log} -type d -mindepth 1 -maxdepth 1  | wc -l
       0
```

Unfortunately cleaning up previous `tmp` directories does not seem like a good idea because they can still be in-use.

### DIFF
--- a/local/php/executor.go
+++ b/local/php/executor.go
@@ -163,7 +163,11 @@ func (e *Executor) DetectScriptDir() (string, error) {
 	return e.scriptDir, nil
 }
 
-// Config determines the right version of PHP depending on the configuration (+ its configuration)
+// Config determines the right version of PHP depending on the configuration
+// (+ its configuration). It also creates some symlinks to ease the integration
+// with underlying tools that could try to run PHP. This is the responsability
+// of the caller to clean those temporary files. One can call
+// CleanupTemporaryDirectories to do so.
 func (e *Executor) Config(loadDotEnv bool) error {
 	// reset environment
 	e.environ = make([]string, 0)
@@ -220,8 +224,10 @@ func (e *Executor) Config(loadDotEnv bool) error {
 	// prepending the PHP directory in the PATH does not work well if the PHP binary is not named "php" (like php7.3 for instance)
 	// in that case, we create a temp directory with a symlink
 	// we also link php-config for pecl to pick up the right one (it is always looks for something called php-config)
-	phpDir := filepath.Join(cliDir, "tmp", xid.New().String(), "bin")
-	e.tempDir = phpDir
+	if e.tempDir == "" {
+		e.tempDir = filepath.Join(cliDir, "tmp", xid.New().String())
+	}
+	phpDir := filepath.Join(e.tempDir, "bin")
 	if err := os.MkdirAll(phpDir, 0755); err != nil {
 		return err
 	}
@@ -284,6 +290,15 @@ func (e *Executor) Config(loadDotEnv bool) error {
 	return err
 }
 
+func (e *Executor) CleanupTemporaryDirectories() {
+	if e.iniDir != "" {
+		os.RemoveAll(e.iniDir)
+	}
+	if e.tempDir != "" {
+		os.RemoveAll(e.tempDir)
+	}
+}
+
 // Find composer depending on the configuration
 func (e *Executor) findComposer(extraBin string) (string, error) {
 	if scriptDir, err := e.DetectScriptDir(); err == nil {
@@ -312,14 +327,7 @@ func (e *Executor) Execute(loadDotEnv bool) int {
 		fmt.Fprintln(os.Stderr, err)
 		return 1
 	}
-	defer func() {
-		if e.iniDir != "" {
-			os.RemoveAll(e.iniDir)
-		}
-		if e.tempDir != "" {
-			os.RemoveAll(e.tempDir)
-		}
-	}()
+	defer e.CleanupTemporaryDirectories()
 	cmd := execCommand(e.Args[0], e.Args[1:]...)
 	environ := append(os.Environ(), e.environ...)
 	gpathname := "PATH"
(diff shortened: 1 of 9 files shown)


local/php/php_builtin_server.go
```diff
@@ -61,9 +60,5 @@ require $script;
 `)
 
 func (p *Server) phpRouterFile() string {
-	path := filepath.Join(p.homeDir, fmt.Sprintf("php/%s-router.php", name(p.projectDir)))
-	if _, err := os.Stat(filepath.Dir(path)); os.IsNotExist(err) {
-		_ = os.MkdirAll(filepath.Dir(path), 0755)
-	}
-	return path
+	return filepath.Join(p.tempDir, fmt.Sprintf("%s-router.php", p.Version.Version))
```
tucksaun, line 63: For review, I moved the routing file to the same location as the FPM config.
This way I can always create and clean the directory no matter the SAPI

local/php/executor.go
```diff
@@ -163,7 +163,11 @@ func (e *Executor) DetectScriptDir() (string, error) {
 	return e.scriptDir, nil
 }
 
-// Config determines the right version of PHP depending on the configuration (+ its configuration)
+// Config determines the right version of PHP depending on the configuration
+// (+ its configuration). It also creates some symlinks to ease the integration
+// with underlying tools that could try to run PHP. This is the responsability
```
fabpot, line 168:
```suggestion
// with underlying tools that could try to run PHP. This is the responsibility
```

fabpot (commented):
Regarding cleaning up existing empty directories, can we at least remove the ones that are older than x days?

tucksaun:
> Regarding cleaning up existing empty directories, can we at least remove the ones that are older than x days?

good idea, will do

tucksaun:
@fabpot cleanup added. I considered directories as stale after 7 days. wdyt?

fabpot:
Great! Thank you.

srsbiz:
Thanks, unfortunately it is somehow not working for me. Are old files removed only on "clean" shutdown of the server? I cannot gracefully stop server with `server:stop` since #397 and while force shutting down with Ctrl+C there is some attempt to remove logs, but it almost always fail
image

tucksaun:
AFAICT this error is not related to this PR and is only related to #397 (on which I didn't had the opportunity to work so far)

srsbiz:
Yes, my question is not about this error, but on conditions required for cleanup process to be performed - if it requires "clean" shutdown by calling `server:stop` or if should be performed on any type of shutdown - like Ctrl+C, stopping process via Task Manager, issuing `kill` command with pid etc.

tucksaun:
On any shutdown. But your error happens during the shutdown process and things are not necessarily linear in Go so your error might end up stopping the cleaning process early on

tucksaun:
@srsbiz see https://github.com/symfony-cli/symfony-cli/pull/569 for an attempt to improve this for you (won't fix the initial error)

PR #569, opened 2025-02-15
refactor: improve cleaning up stale temp directories if workers fail to stop

tucksaun:
I'm not entirely sure about this one. wdyt @fabpot ?


srsbiz:
That change deleted old files, but errored on each dir it emptied (could not access XXX it is used by other process)
On second run and close no empty dir was deleted and no errors

PR #576, opened 2025-02-18
fix: fix generated PSH related code format

tucksaun:


local/platformsh/generator/commands.go
```diff
@@ -93,7 +94,11 @@ func generateCommands() {
 	if err != nil {
 		panic(err)
 	}
-	f.Write(buf.Bytes())
+	source, err := format.Source(buf.Bytes())
```
fabpot, line 97: TIL

PR #579, opened 2025-02-22
Improve warning message when Composer is not found

tucksaun:
refs #424

Possible outputs:

```
  WARNING: No Composer installation found.
  Downloading Composer for you, but it is recommended to install Composer yourself, instructins available at https://getcomposer.org/download/
  (running /Users/tucksaun/.symfony5/composer/composer.phar --version)

Composer version 2.8.3 2024-11-17 13:13:04
PHP version 8.2.27 (/opt/homebrew/Cellar/php@8.2/8.2.27/bin/php)
Run the "diagnose" command to get more detailed diagnostics output.
```

```
  WARNING: Detected Composer file (/Users/tucksaun/Work/src/github.com/symfony-cli/symfony-cli/composer.phar) is not a valid PHAR or PHP script.
  Downloading Composer for you, but it is recommended to install Composer yourself, instructins available at https://getcomposer.org/download/
  (running /Users/tucksaun/.symfony5/composer/composer.phar --version)

Composer version 2.8.3 2024-11-17 13:13:04
PHP version 8.2.27 (/opt/homebrew/Cellar/php@8.2/8.2.27/bin/php)
Run the "diagnose" command to get more detailed diagnostics output.
```

PR #581, opened 2025-02-25
feat: allow to define where is Composer using SYMFONY_COMPOSER_PATH env var

### LINKED ISSUE #424: Composer detection problem if composer is wrapped in a shell script
os: debian12
symfony: 5.8.6
composer: 2.6.6

my composer file is a shell script which is doing different stuff for me before actually running composer. so i there is the bash script `/usr/local/bin/composer`  which runs the actual composer file `/usr/local/bin/composer-bin`

shouldn't a simple detection be enough?
```
command -v composer
/usr/local/bin/composer

which composer
/usr/local/bin/composer
```

tucksaun:
This is my take on supporting Composer not being directly executed by Symfony CLI.

To be noted: by doing so, there is no guarantee that the correct PHP version will be executed as the program being executed is solely responsible for what is going on after. However we keep your "virtual" PATH technic in place so if it uses the PATH, then it should work most of the time.

Closes #424
Supersedes and closes #579

### DIFF
--- a/local/php/executor.go
+++ b/local/php/executor.go
@@ -74,6 +74,10 @@ func GetBinaryNames() []string {
 	return []string{"php", "pecl", "pear", "php-fpm", "php-cgi", "php-config", "phpdbg", "phpize"}
 }
 
+func (e Executor) CommandLine() string {
+	return strings.TrimSpace(strings.Join(e.Args, " "))
+}
+
 func (e *Executor) lookupPHP(cliDir string, forceReload bool) (*phpstore.Version, string, bool, error) {
 	phpStore := phpstore.New(cliDir, forceReload, nil)
 	v, source, warning, err := phpStore.BestVersionForDir(e.scriptDir)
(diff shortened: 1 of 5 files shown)


local/php/executor.go
```diff
@@ -74,6 +74,10 @@ func GetBinaryNames() []string {
 	return []string{"php", "pecl", "pear", "php-fpm", "php-cgi", "php-config", "phpdbg", "phpize"}
 }
 
+func (e Executor) CommandLine() string {
```
fabpot, line 77: Nice one!

fabpot (approved):

fabpot:
Thank you @tucksaun Would it make sense to have a note somewhere in the docs (or maybe this is an edge case)?

tucksaun:
> Would it make sense to have a note somewhere in the docs (or maybe this is an edge case)?

To me this looks mostly as an edge case: we only have a handful of users impacted so far (2 publicly and I might remember one or two mentioning it to me privately) and so far the only platform impacted was nix (because of the shell wrapper)

PR #584, opened 2025-03-13
Enhancement: manual IP specification during `server:ca:install` command

ramity:
Hi, all. 

I hit a bit of an admittedly fringe issue during some dev work and needed to be able to add a local IP to the list of ips added to the cert generated during the `server:ca:install` go script. I extended the go script with an additional StringFlag with a default value matching the current hardcoded value to preserve the current functionality.

### DIFF
--- a/commands/local_server_ca_install.go
+++ b/commands/local_server_ca_install.go
@@ -78,7 +78,7 @@ var localServerCAInstallCmd = &console.Command{
 			ips := strings.Split(c.String("ips"), ",")
 			err := ca.MakeCert(p12, ips)
 			if err != nil {
-				return errors.Wrap(err, "failed to generate a certificate")
+				return errors.Wrap(err, "failed to generate a default certificate")
 			}
 		}
 
--- a/commands/local_server_ca_install.go
+++ b/commands/local_server_ca_install.go
@@ -39,7 +39,7 @@ var localServerCAInstallCmd = &console.Command{
 	Flags: []console.Flag{
 		&console.BoolFlag{Name: "renew", Usage: "Force generating a new CA"},
 		&console.BoolFlag{Name: "force", Aliases: []string{"f"}, Usage: "Force reinstalling current CA"},
-		&console.StringFlag{Name: "ips", Usage: "Comma-separated list of IPs for the certificate", Value: "localhost,127.0.0.1,::1"},
+		&console.StringFlag{Name: "ips", Usage: "Comma-separated list of IPs for the certificate", DefaultText: "localhost,127.0.0.1,::1"},
 	},
 	Action: func(c *console.Context) error {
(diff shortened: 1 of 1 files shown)


ramity:
Just got around to testing locally. I downloaded the most recent release artifact from my PR locally and instead of running the following in my dockerfile:

```
RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | bash
RUN apt install -y symfony-cli
```

I substituted it with:

```
COPY data/core-backend/custom-symfony-binary.zip /tmp/custom-symfony-binary.zip
RUN cd /tmp && unzip /tmp/custom-symfony-binary.zip -d /tmp
RUN cd /tmp && tar -xvzf /tmp/symfony-cli_linux_amd64.tar.gz
RUN mv /tmp/symfony /usr/local/bin/symfony
```

With the custom binary installed, I ran:

```
symfony server:ca:install --ips="192.168.1.4,localhost,127.0.0.1,::1"
```

Which produced a certificate with the 4 IPs listed correctly:

image

I tested without the ips flag specified and got an error, so marking as draft while I figure that out.

ramity:
## Ready for review but could use an additional test, see remaining section below

I resolved the issue and repeated my build process to generate a cert without the flag being set and produced the following cert:

image

## Use case

My use case for this mostly stems from using a separate devices to host and access my dev environments. I'm exploring progressive web app related work, so the baked in path for https in a single command was very nice. Only issue is that accessing the site from an external client instead of localhost will generate a cert error since it's not being served over a valid ip. With this PR, I can add the additional IP to resolves the cert error to simply a cert authority error, which can be resolved with adding the cert locally.

I'm unsure if this functionality is really applicable to many or best practice. Some conversation on that wouldn't hurt, but this achieves what I need it to.

## Remaining

I'm a little unfamiliar with go, but if someone could point me to where the `server:ca:install` command is being tested during the release runner, I could add an additional test case to test with the ips flag set.

ramity:
The fruits of my labor - a valid service worker served over https from another computer local on the network running a docker container running symfony server.

image

tucksaun:
I’m not sure I would recommend adding the certificates we generate on another system that the local system where it has been generated.

Also, the `default.p12` certificate is used by default for all instances of the Symfony CLI so removing `localhost` and `127.0.0.1` can lead to unexpected issues.

Additionally, the remote IP can change at any time, with this implementation it means we need to make possible to regenerate it when needed.

Finally, supporting this in `server:ca:install` means it also needs to be supported in `local:server:start`.

IF we follow this path I think I would rather generate a certificate on the fly based on the listenIP to avoid any conflicts or user errors.

However, I might have a solution for you: the HTTPS proxy usage seems to cover your use case as it’s reachable on other IPs and the certificate generation based on DNS names means no issue with IP change. would that work for you?

fabpot:
Closing as no more activity. Feel free to comment.

PR #587, opened 2025-03-23
Add support for `pie`

alexandre-daubois:
Just like `composer` and `pecl`, I think the `symfony` tool could propose the `pie` subcommand.

For now I don't think we can get the phar signature of the latest release. Could this be possible to maybe include the signature as a release artifact maybe @asgrim, or similar mechanism? 🙂 Composer offers to download the latest signature at https://composer.github.io/installer.sig. Please tell me if I can be of any help on this!

This could also be an opportunity to resume #435 by adding a few integration tests for this command?

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -52,6 +52,7 @@ func CommonCommands() []*console.Command {
 	localCommands := []*console.Command{
 		binConsoleWrapper,
 		composerWrapper,
+		pieWrapper,
 		phpWrapper,
 		bookCheckReqsCmd,
 		bookCheckoutCmd,
@@ -179,6 +180,7 @@ Environment variables to use Platform.sh/Upsun relationships or Docker services 
 {{with .Command "composer"}}  <info>{{.PreferredName}}</>{{"\t"}}{{.Usage}}{{end}}
 {{with .Command "console"}}  <info>{{.PreferredName}}</>{{"\t"}}{{.Usage}}{{end}}
 {{with .Command "php"}}  <info>{{.PreferredName}}</>{{"\t"}}{{.Usage}}{{end}}
+{{with .Command "pie"}}  <info>{{.PreferredName}}</>{{"\t"}}{{.Usage}}{{end}}
 
 `
 }
--- a/local/php/pie.go
+++ b/local/php/pie.go
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2025-present Fabien Potencier <fabien@symfony.com>
+ *
+ * This file is part of Symfony CLI project
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package php
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"github.com/symfony-cli/symfony-cli/util"
+)
+
+type PieResult struct {
+	code  int
+	error error
+}
+
+func (p PieResult) Error() string {
+	if p.error != nil {
+		return p.error.Error()
+	}
+
+	return ""
+}
+
+func (p PieResult) ExitCode() int {
+	return p.code
+}
+
+func Pie(dir string, args, env []string, stdout, stderr, logger io.Writer, debugLogger zerolog.Logger) PieResult {
+	e := &Executor{
+		Dir:        dir,
+		BinName:    "php",
+		Stdout:     stdout,
+		Stderr:     stderr,
+		SkipNbArgs: -1,
+		ExtraEnv:   env,
+		Logger:     debugLogger,
+	}
+
+	if piePath := os.Getenv("SYMFONY_PIE_PATH"); piePath != "" {
+		debugLogger.Debug().Str("SYMFONY_PIE_PATH", piePath).Msg("SYMFONY_PIE_PATH has been defined. User is taking control over PIE detection and execution.")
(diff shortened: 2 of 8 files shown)


local/php/assert.go
```diff
@@ -0,0 +1,30 @@
+package php
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+)
+
+// assertIsPHPScript checks that the provided file is indeed a phar/PHP script (not a .bat file)
+func assertIsPHPScript(path string) bool {
```
tucksaun, line 10: assert might seems a bit strong for developers coming from PHP and they might think this would stop the execution if the condition was not met.
WDYT renaming this function to `checkIsPHPScript` instead?

local/php/pie.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"github.com/symfony-cli/symfony-cli/util"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
```
tucksaun, line 31: in the Go ecosystem, one usually splits STL packages from other third-party ones:
```suggestion
	"fmt"
	"io"
	"net/http"
	"os"
	"path/filepath"
	"strings"
	
	"github.com/pkg/errors"
	"github.com/rs/zerolog"
	"github.com/symfony-cli/symfony-cli/util"
```

local/php/pie.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func findPie(logger zerolog.Logger) (string, error) {
+	for _, file := range []string{"pie", "pie.phar"} {
+		logger.Debug().Str("source", "PIE").Msgf(`Looking for PIE in the PATH as "%s"`, file)
+		if pharPath, _ := LookPath(file); pharPath != "" {
+			// On Windows, we don't want the .bat, but the real pie phar/PHP file
+			if strings.HasSuffix(pharPath, ".bat") {
+				pharPath = pharPath[:len(pharPath)-4] + ".phar"
+			}
```
tucksaun, line 103: I’m not familiar with PIE but I don’t recall a bat wrapper being shipped with PIE.
Do you have more info about this aspect?

tucksaun:
> For now I don't think we can get the phar signature of the latest release. Could this be possible to maybe include the signature as a release artifact maybe @asgrim, or similar mechanism? 🙂

Seems like we should be able by reimplementing the logic behind `gh attestation verify`. (for reference here is a link to start the investigation: https://github.com/cli/cli/blob/trunk/pkg/cmd/attestation/verify/verify.go). though having a signature as Composer would be definitely easier to implement and maintain.

tucksaun (commented):
Thanks for this solid addition!

Having not play with PIE yet, I would just like give it a go locally before a definitive approval 🙂

local/php/pie.go
alexandre-daubois: You're right, it's an artifact from the Composer wrapper. I removed it

local/php/assert.go
alexandre-daubois: Suits me fine!

commands/root.go
```diff
@@ -143,6 +144,7 @@ func WelcomeAction(c *console.Context) error {
 		localServerStopCmd,
 		localSecurityCheckCmd,
 		composerWrapper,
+		pieWrapper,
```
fabpot, line 147: I don't think it deserves to be in this list, which should be kept as short as possible.

local/php/assert.go
```diff
@@ -0,0 +1,30 @@
+package php
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+)
+
+// checkIsPHPScript checks that the provided file is indeed a phar/PHP script (not a .bat file)
+func checkIsPHPScript(path string) bool {
```
fabpot, line 10: Maybe just:

```suggestion
func isPHPScript(path string) bool {
```

local/php/assert_test.go
```diff
@@ -25,24 +25,24 @@ import (
 	. "gopkg.in/check.v1"
 )
 
-type ComposerSuite struct{}
+type AssertSuite struct{}
```
fabpot, line 28: Let's find another name. Maybe `utils.go` would be better.

local/php/assert.go
fabpot: I realize now that this was the old name... which I like better :)

local/php/pie.go
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (c) 2021-present Fabien Potencier <fabien@symfony.com>
```
fabpot, line 2:
```suggestion
 * Copyright (c) 2025-present Fabien Potencier <fabien@symfony.com>
```

fabpot (changes requested):

commands/root.go
alexandre-daubois: This means that it won't be shown in the available wrappers when running `symfony -h`, I'm afraid it makes it undiscoverable?

alexandre-daubois: Removed from this list and added to the list of available wrappers when using `-h`

local/php/assert_test.go
alexandre-daubois: Suits me fine

fabpot (approved):

tucksaun:
> I'll have a look at what's possible about `gh attestation verify` when I have some time 🙂

PIE is quite cutting-edge as of now so I think we can merge as-is and address this in the coming days/weeks.

tucksaun (approved):

tucksaun:
Thank you @alexandre-daubois !

OskarStark:
Can we have any docs ￼

alexandre-daubois:
I'm not sure where we should add this. I would have go for https://symfony.com/doc/current/setup/symfony_server.html, but "Symfony Local Web Server" doesn't seem suitable. However it seems the only place we could add this type of doc? Maybe a whole section to wrappers should be added also.

tucksaun:
> I'm not sure where we should add this. I would have go for https://symfony.com/doc/current/setup/symfony_server.html, but "Symfony Local Web Server" doesn't seem suitable. However it seems the only place we could add this type of doc? Maybe a whole section to wrappers should be added also.

I started to talk about this with a couple of people and I believe we should introduce a dedicated documentation for the CLI with the more advanced options and only have an introduction in the Framework doc (with a link to the dedicated one).

This could also help with the fact that currently the documentation for the CLI does not take the CLI version into account because it is being embedded in the Framework documentation.

WDYT?

alexandre-daubois:
100% agree, I'm pretty sure many (most?) people underuse Symfony CLI because there's no clear dedicated page to it. Let's create an issue in symfony-docs 🙂

PR #597, opened 2025-04-10
Add PIE to the 'Work on a project locally' help section

tucksaun:
Screenshot 2025-04-10 at 15 30 26

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -145,6 +145,7 @@ func WelcomeAction(c *console.Context) error {
 		localSecurityCheckCmd,
 		composerWrapper,
 		binConsoleWrapper,
+		pieWrapper,
 		phpWrapper,
 	})
 	terminal.Println("")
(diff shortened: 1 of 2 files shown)


fabpot:
Would it makes sense to add it to the `phpWrapper` line below? That would bemore consistent from an end user perspective.

tucksaun:
> Would it makes sense to add it to the `phpWrapper` line below? That would bemore consistent from an end user perspective.

This is something I considered but PIE is kind of in a mixed situation because it's close to `composer` situation but impact the PHP setup 🤔 
Code wise it's also "special" as all possible PHP binary names (and so wrappers) are handled somewhere where we can't add `pie` because it would mess the `pie` support.

So updated the PR to move towards your suggestion but we can easily revert it if we want to.

fabpot:
I'm wondering if we should remove `pear` support.

tucksaun:
> I'm wondering if we should remove `pear` support.

I don't think I ever used (except during initial implementation)

PR #611, opened 2025-05-12
feat: add support for `.symfony.local.dist.yaml` and `.symfony.local.override.yaml` config files

### LINKED ISSUE #606: [RFC] Look for .dist configuration files ?
AFAIK, the current CLI does not handle ".dist" files:  `.symfony.dist.yaml` --which might not be that useful--, nor ` .symfony.dist.local.yaml`, `.symfony.local.dist.yaml`, or `.symfony.local.yaml.dist`  (feels like every tool has its own conventions 😅).

In a project, it makes sense to share "standard" development configurations in the team. 

However, I find it quite unnecessary and resource-consuming to run multiple Node watchers and Tailwind processes when I just want to work on PHP.

I know there's a `--no-worker` option, but having a way to define personal local configurations that are not shared in VCS would be _fantastic_.  🙏  

I can't help much with the implementation (but I can make a solid 🦆), but I can definitely bring you some coffee next week. :))

tucksaun:
fix #606

still WIP but opening early so that one can try it locally and that we can discuss some details

### DIFF
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -60,32 +60,11 @@ var localServerStartCmd = &console.Command{
 	Aliases:     []*console.Alias{{Name: "server:start"}, {Name: "serve"}},
 	Usage:       "Run a local web server",
 	Description: localWebServerProdWarningMsg,
-	Flags: []console.Flag{
+	Flags: append(
+		project.ConfigurationFlags,
 		dirFlag,
-		&console.BoolFlag{Name: "allow-http", Usage: "Prevent auto-redirection from HTTP to HTTPS"},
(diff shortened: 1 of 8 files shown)


Kocal:
What's the file `.symfony.local.override.yaml`?

tucksaun:
> What's the file `.symfony.local.override.yaml` for?

While implementing the feature, I figured similar tools (Docker Compose for instance) also uses an `.override.yaml` pattern and because support it is only a matter of adding a loop iteration I added. This way, one can choose what they prefer.

smnandre:
tested: works perfectly for me! Thanks a lot @tucksaun

tucksaun:
I think we are good 🎉 
But this is quite some change so please review and test thoroughly 🙏

fabpot (approved):

PR #616, opened 2025-05-16
prefer os specific config dirs

### LINKED ISSUE #133: Use system config folder to store files
Hey,

the CLI currently uses `$HOME/.symfony5` to store files. It would be nice if the cli could use the system folder like XDG_CONFIG_HOME on Linux and `Library/Application Support/` on macOS

tjblackheart:
Closes #133

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -20,28 +20,51 @@
 package util
 
 import (
+	"os"
 	"os/user"
 	"path/filepath"
-
-	"github.com/mitchellh/go-homedir"
+	"runtime"
 )
 
 func GetHomeDir() string {
-	return filepath.Join(getUserHomeDir(), ".symfony5")
+	return getUserHomeDir()
 }
 
 func getUserHomeDir() string {
+	dir := "symfony5"
+	dotDir := "." + dir
+
 	if InCloud() {
 		u, err := user.Current()
 		if err != nil {
-			return "/tmp"
+			return filepath.Join(os.TempDir(), dir)
 		}
-		return "/tmp/" + u.Username
+		return filepath.Join(os.TempDir(), u.Username, dir)
 	}
 
-	if homeDir, err := homedir.Dir(); err == nil {
-		return homeDir
+	home, err := os.UserHomeDir()
+	if err != nil {
+		home = "."
 	}
 
-	return "."
+	// use the old path if it exists already
+	fallback := filepath.Join(home, dotDir)
+	if _, err := os.Stat(fallback); !os.IsNotExist(err) {
+		return fallback
+	}
+
+	// macos only: if $HOME/.config exist, prefer that over 'Library/Application Support'
+	if runtime.GOOS == "darwin" {
+		dotconf := filepath.Join(home, ".config")
+		if _, err := os.Stat(dotconf); !os.IsNotExist(err) {
+			return filepath.Join(dotconf, dir)
+		}
+	}
+
+	userCfg, err := os.UserConfigDir()
+	if err != nil {
+		return fallback
+	}
+
+	return filepath.Join(userCfg, dir)
 }


shyim (approved):

util/util.go
```diff
@@ -20,28 +20,51 @@
...
+	"os"
 	"os/user"
 	"path/filepath"
-
-	"github.com/mitchellh/go-homedir"
+	"runtime"
 )
 
 func GetHomeDir() string {
-	return filepath.Join(getUserHomeDir(), ".symfony5")
```
tucksaun, line 30: Wondering if we should not move to string concatenation to be able to keep the symfony folder name in a single place 🤔

util/util.go
```diff
@@ -20,28 +20,51 @@
...
+	return getUserHomeDir()
 }
 
 func getUserHomeDir() string {
+	dir := "symfony5"
+	dotDir := "." + dir
```
tucksaun, line 35: Used only once so we can inline it

util/util.go
```diff
@@ -20,28 +20,51 @@
...
+		return filepath.Join(os.TempDir(), u.Username, dir)
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		home = "."
+	}
+
+	// use the old path if it exists already
+	fallback := filepath.Join(home, dotDir)
```
tucksaun, line 51: I would name this "legacy" instead of "fallback"

util/util.go
```diff
@@ -20,28 +20,51 @@
...
+			return filepath.Join(dotconf, dir)
 		}
-		return "/tmp/" + u.Username
 	}
 
-	if homeDir, err := homedir.Dir(); err == nil {
-		return homeDir
+	userCfg, err := os.UserConfigDir()
+	if err != nil {
+		return fallback
```
tucksaun, line 66: I would continue fallbacking to `.`

util/util.go
```diff
@@ -20,28 +20,51 @@
...
 		if err != nil {
-			return "/tmp"
+			return filepath.Join(os.TempDir(), dir)
+		}
+		return filepath.Join(os.TempDir(), u.Username, dir)
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		home = "."
```
tucksaun, line 47: Given my last comment it would be "covered"/"used" as the latest fallback so I would not use it here (inverting the condition and use the legacy only if present)

tucksaun (commented):
Thank you for starting the work on this

util/util.go
```diff
@@ -20,28 +20,51 @@
...
+	"runtime"
 )
 
 func GetHomeDir() string {
-	return filepath.Join(getUserHomeDir(), ".symfony5")
+	return getUserHomeDir()
 }
 
 func getUserHomeDir() string {
+	dir := "symfony5"
```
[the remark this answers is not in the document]
tjblackheart: good point.

util/util.go
tjblackheart: I like the "const" approach, as it is more clear than this.

util/util.go
tjblackheart: I don't understand. "home" here is just a fallback in case anything goes wrong with the detection?

tjblackheart: ... a more proper way would be to use os.Getwd() I guess

tucksaun: my point is that we have to handle a global fallback case if everything fails.
this is the `.` case.
because we will handle it as a last resort anyway, we don't have to deal with it here, which allows for early returns without variables leaking or inaccurate naming (the naming is inaccurate in the last `return legacy` in the sense it is a fallback not necessarily a legacy path). eg, something as:

```go
// use the old path if it exists already
if home, err := os.UserHomeDir(); err != nil {
	legacyPath := filepath.Join(home, "."+confDir)
        if _, err := os.Stat(legacyPath); !os.IsNotExist(err) {
		return legacyPath
	}
}
```

util/util.go
```diff
@@ -20,28 +20,50 @@
...
+
+	// use the old path if it exists already
+	legacy := filepath.Join(home, "."+confDir)
+	if _, err := os.Stat(legacy); !os.IsNotExist(err) {
+		return legacy
+	}
+
+	// macos only: if $HOME/.config exist, prefer that over 'Library/Application Support'
+	if runtime.GOOS == "darwin" {
+		dotconf := filepath.Join(home, ".config")
```
tucksaun, line 58: this check is not strict enough and might make the `$HOME/.config` directory be used without the user "consent". If an application creates a directory at any point, the CLI will start to use the previous directory without the user knowing it.

util/util.go
tjblackheart: I see, thanks for the explanation. "home" at this point was the base for the following legacy folder check and also the check for a macos `~/.config`. I guess if we remove the macos stuff this can go too.

tjblackheart: Hey @tucksaun, you think we can finish up this little thing? If these mere 10 lines of code are somehow not up to your standard please say so and I just close it, no problem. If not, I don't see whats the hold up. Thanks!

tucksaun: > If we include this like your proposal though, legacyPath will be needed again in the last return.

No because the last return (acting as a fallback) would then be `return "./." + confDir`.

Overall this organisation is mostly nice for us maintainers because it means the `if` block handling the legacy path is only responsible for handling legacy, not the fallback:
- we can have a warning about the use of the legacy path and the need to migrate
- we only have to remove this block when we remove support for legacy config directories

tucksaun:
I took the liberty to rebase the PR, update the new directory named and add a warning to let users know about the change

CI errors seems unrelated (I guess the GH action has a hard time figuring out was is new from what is not because the PR has been opened before it was added)

util/homedir.go
```diff
@@ -0,0 +1,70 @@
...
+		configurationPath = filepath.Join(userCfg, ConfigurationDirectory)
+	}
+
+	// use the legacy path if it exists already
+	if home, err := os.UserHomeDir(); err == nil {
+		legacyPath := filepath.Join(home, LegacyConfigurationDirectory)
+		if _, err := os.Stat(legacyPath); !errors.Is(err, fs.ErrNotExist) {
+			terminal.Logger.Warn().Str("directory", legacyPath).Err(err).Msg("Legacy configuration directory detected")
+			legacyPathWarning.Do(func() {
+				terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin).Warning(fmt.Sprintf("The configuration location for the Symfony CLI has changed in v5.12.0.\nHaving the configuration stored in \"$HOME/.symfony5\" is deprecated and will not be supported in next major versions.Please migrate the \"%s\" directory manually to \"%s\" at your earliest convenience after stopping the proxy and every instances running.", legacyPath, configurationPath))
```
fabpot:
```suggestion
				terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin).Warning(fmt.Sprintf("The configuration location for the Symfony CLI has changed in v5.16.0.\nHaving the configuration stored in \"$HOME/.symfony5\" is deprecated and will not be supported in next major versions.Please migrate the \"%s\" directory manually to \"%s\" at your earliest convenience after stopping the proxy and every instances running.", legacyPath, configurationPath))
```

util/homedir.go
```diff
@@ -0,0 +1,70 @@
...
+	} else {
+		configurationPath = filepath.Join(userCfg, ConfigurationDirectory)
+	}
+
+	// use the legacy path if it exists already
+	if home, err := os.UserHomeDir(); err == nil {
+		legacyPath := filepath.Join(home, LegacyConfigurationDirectory)
+		if _, err := os.Stat(legacyPath); !errors.Is(err, fs.ErrNotExist) {
+			terminal.Logger.Warn().Str("directory", legacyPath).Err(err).Msg("Legacy configuration directory detected")
+			legacyPathWarning.Do(func() {
```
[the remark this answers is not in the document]
tucksaun: Honestly I don't remember.

I tried to found some migration code in earlier versions and I couldn't.

The problems I see with renaming automatically are:
- renaming won't work across filesystems boundaries
- copying or moving individual files mike take some times because of logs and PID states
- we need to be sure all instances and the proxy are stopped otherwise some FD for previous files might still be hold by other processes.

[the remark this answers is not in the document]
tucksaun: If determining the best cours of action here is too much of a hassle I'm fine with extracting it from the current PR into another PR so that we can discuss it without any rush

util/homedir.go
```diff
@@ -0,0 +1,70 @@
...
+				terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin).Warning(fmt.Sprintf("The configuration location for the Symfony CLI has changed in v5.16.0.\nHaving the configuration stored in \"$HOME/.symfony5\" is deprecated and will not be supported in next major versions.Please migrate the \"%s\" directory manually to \"%s\" at your earliest convenience after stopping the proxy and every instances running.", legacyPath, configurationPath))
```
fabpot:
```suggestion
				terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin).Warning(fmt.Sprintf("The configuration location for the Symfony CLI has changed in v5.16.0.\nHaving the configuration stored in \"$HOME/.symfony5\" is deprecated and will not be supported in the next major version. Please move the \"%s\" directory manually to \"%s\" after stopping the proxy and all running web servers.", legacyPath, configurationPath))
```

fabpot (approved):

fabpot:
Thank you @tjblackheart.

PR #622, opened 2025-06-11
Fix CLI info injection into Symfony 7.3+ WDT, fix #620

### LINKED ISSUE #620: Symfony WDT does not include Symfony CLI information since Symfony 7.3
Hi,

Sorry if it's not the right repository where to report this bug... A 1 in 2 chance 😅 

Soooo, I'm upgrading all my Symfony apps since Friday, and noticed something was wrong but didn't know what... until today.

After upgrading a Symfony application to 7.3, the debug toolbar item about Symfony CLI (with information about Symfony CLI version, Docker etc) does not appear anymore.

tucksaun:


fabpot:
Thank you @tucksaun

PR #623, opened 2025-06-11
chore: migrate Goreleaser brews (deprecated) to homebrew_casks

tucksaun:
In recent PRs I spotted this, so I guess we should migrate:

Screenshot 2025-06-11 at 16 18 09

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -87,7 +87,7 @@ release:
   footer: |
     **Full Changelog**: https://github.com/symfony-cli/symfony-cli/compare/{{ .PreviousTag }}...{{ .Tag }}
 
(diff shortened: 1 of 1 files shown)


fabpot:
Tests don't pass ;)

tucksaun:
Let's forget about this one for now because there's several drawbacks:
- the GOARM selection does not seem possible (the doc mention only one GOARM version is supported but not how to select the one we want to publis)
- We can define dependencies but only casks or formulas and we can't make them optional
- It seems the service installation is now done by copying a definition file but it's unclear how to create the file

So all in all it will require more work and time that "simply replacing one with the other" as the doc states 😅

PR #626, opened 2025-06-12
do not truncate log files on each LogWriter calls

tucksaun:
We used to have a single concurrent usage of the `LogWriter`, but this is not the case anymore since b29665f7afcba0a71a30077d3516aa6076908924. Because of this, we can lose some logs (the runner ones for instance) as the file is being truncated during the `buildCmd` call when the `LogWrite` is fetched.

To fix this, I propose to instantiate the `LogWriter` once, which will effectively truncate the file only once per `server:start` launch.

### DIFF
--- a/local/pid/pidfile.go
+++ b/local/pid/pidfile.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


local/pid/pidfile.go
```diff
@@ -258,15 +259,24 @@ func (p *PidFile) LogReader() (io.ReadCloser, error) {
 }
 
 func (p *PidFile) LogWriter() (io.WriteCloser, error) {
+	// instantiate the log writer only once per process lifetime, this is useful
+	// to have a single truncate (and thus a clean log file) at the beginning of
+	// the process management but not truncate the log file when the process is
+	// restarted.
+	if p.lw != nil {
```
fabpot, line 266: Does it need to be guarded by a sync.Once?

tucksaun: done but moving to draft because I didn't had the opportunity to run it locally nor run the test suite

tucksaun: just tried it locally and worked as expected

fabpot:
Thank you @tucksaun

PR #634, opened 2025-07-11
add Access-Control-Expose-Headers to CORS headers

fnash:
Same way allowing HTTP clients to send cross-origin requests (using --allow-cors option), this PR allows scripts in the browser access custom headers sent by the server.

Custom headers are made available when server response contains `Access-Control-Expose-Headers` header.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Expose-Headers

### DIFF
--- a/local/http/cors.go
+++ b/local/http/cors.go
@@ -26,7 +26,7 @@ import (
 )
 
 func corsWrapper(h http.Handler, logger zerolog.Logger) http.Handler {
-	var corsHeaders = []string{"Access-Control-Allow-Origin", "Access-Control-Allow-Methods", "Access-Control-Allow-Headers"}
+	var corsHeaders = []string{"Access-Control-Allow-Origin", "Access-Control-Allow-Methods", "Access-Control-Allow-Headers", "Access-Control-Expose-Headers"}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		for _, corsHeader := range corsHeaders {


fnash:
@tucksaun If Symfony server stays for local development so yes this facility makes sense IMO the same way the server allows all http verbs and all origins.

Some applications need to send custom headers to the client, and in this context javascript needs to access those response headers.

In production hosted environments, it is not the same context. Symfony server should not be used. And if there is CORS need, granularity should be defined in exposed headers as well as http verbs and allowed origins for security reasons.

tucksaun:
> Some applications need to send custom headers to the client, and in this context javascript needs to access those response headers.

Yes I agree but if you send custom headers, this is done in the backend code and as such one can control the CORS headers. The `--allow-cors` option is really meant to be used for static assets which usually do not need custom headers access.

fnash:
> Yes I agree but if you send custom headers, this is done in the backend code and as such one can control the CORS headers. 

Backend code is not meant to deal with development environment. It is not part of the application nor the production environment unfortunately. 😞 

I am working on an application where front and back are in the same domain. No CORS issue.
Using traefik for example in dev env with same host could resolve it. No code.
If I choose to use symfony server (localhost:8000) and npm start (localhost:3000), I need to enable cors to make it work.


> The --allow-cors option is really meant to be used for static assets which usually do not need custom headers access.

Indeed. True for static assets.
But if it is really meant to be used for assets, why exposing all HTTP verbs? Same for headers in response to prefilghts requests.  🤔

Kocal:
I have no strong opinions on this, and I'm far from being a CORS expert, but since Symfony CLI should only be used for dev purpose, I think it's fine here (and so, even if the backend code could add the header too)

PR #652, opened 2025-09-23
Remove project:init to use the default Upsun one

fabpot:
Before merging, we need to make sure that the Upsun AI-powered `project:init` command works well for Symfony projects.


chirimoya:
@fabpot quick question — will this change influence the PR for adding Sulu support? #645

PR #655, opened 2025-09-25
Git is not requirement for the `new` command

fabpot:


commands/local_new.go
```diff
@@ -168,10 +167,8 @@ var localNewCmd = &console.Command{
 		}
 
 		if !c.Bool("no-git") {
-			if _, err := exec.LookPath("git"); err == nil {
-				if err := initProjectGit(c, dir); err != nil {
-					return err
-				}
+			if err := initProjectGit(c, dir); err != nil {
```
fabpot, line 170: At this point, the check has already been done, so we know Git is available.

commands/local_new.go
```diff
@@ -249,6 +248,9 @@ func initCloud(c *console.Context, brand platformsh.CloudBrand, minorPHPVersion,
 		return err
 	}
 
+	if !c.Bool("no-git") {
```
chirimoya, line 251: @fabpot this early return prevents adding the cloud files to git and breaks the cloud initialization. I think it should be changed to achieve the intended behavior:

```
if c.Bool("no-git") {
    return nil
}
```

Issue #639, opened 2025-08-03
Do not display previous logs when starting the server


Kocal:
I've always found it annoying to not being to see on what URLs your project is available, because of _tons_ of logs you see when booting the server, each time I have to scroll-up to know what URLs to use:

https://github.com/user-attachments/assets/42907a57-8d64-4904-b17b-3f842c446b69

When booting a server, I don't really care about logs from a previous `symfony serve`, the most important things must be visible:
1. the two different warnings about "The local web server is optimized for local development..." and "that the Symfony CLI only listens on 127.0.0.1 by default"
2. the URLs where I can access my app

Do you think never displaying logs from a previous session could be possible?

Thanks!

tucksaun:
Not sure it would be really easy because of the asynchronous nature of this part of the code 🤔 

In the meantime, would the usage of `symfony open:local` help you in your use case?

Kocal:
> In the meantime, would the usage of `symfony open:local` help you in your use case?

Yes it could, but I usually run the web-server through `symfony serve`, in foreground (allowing me to easily kill it without having to type another command), which would requires to open a new terminal and type `symfony open:local`.

> Additionally, we tail 10 lines on existing logs because when we start the tailing process, the process we want to monitor might have already started writing logs that we need to show.

That's a fair point. But since logs have a date-time, maybe we could not display logs written before running the command? If I look back to my video, I can see ~20 lines that are not useful and could be removed: 

<img width="1671" height="471" alt="Image" src="https://github.com/user-attachments/assets/b355b2ec-42a8-477a-8c17-19fb05d247c9" />

tucksaun:
That makes me think: what about introducing a `—open` flag to
`server:start`  that would open the URL once the server started?

tucksaun:
We should already have everything we need thanks to  the `open:local`
command

PR #668, opened 2025-11-11
Add `--open` flag to `server:start` command

Kocal:
Related to https://github.com/symfony-cli/symfony-cli/issues/639

Passing the flag `--open` to `symfony serve` or `symfony server:start` open the first reachable URL in the default web-browser.

### DIFF
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -64,6 +64,7 @@ var localServerStartCmd = &console.Command{
 		project.ConfigurationFlags,
 		dirFlag,
 		&console.BoolFlag{Name: "no-humanize", Usage: "Do not format JSON logs"},
+		&console.BoolFlag{Name: "open", Usage: "Open the application in the browser after starting the server"},
 	),
 	Action: func(c *console.Context) error {
 		ui := terminal.SymfonyStyle(terminal.Stdout, terminal.Stdin)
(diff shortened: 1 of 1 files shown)


tucksaun:
I think I agree with @fabpot here: while I see the use of the `--open` flag, I'm not sure having it the configuration is that useful

commands/openers.go
```diff
@@ -143,16 +142,7 @@ var projectLocalServiceOpenCmd = &console.Command{
...
 }
-
-func abstractOpenCmd(url string) {
-	if err := open.Run(url); err != nil {
-		terminal.Eprintln("<error>Error while opening:", err, "</>")
-		terminal.Eprintfln("Please visit <href=%s>%s</> manually.", url, url)
-	} else {
-		terminal.Eprintfln("Opened: <href=%s>%s</>", url, url)
-	}
-}
```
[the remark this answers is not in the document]
Kocal: I wasn't sure where to put it, so I simply reverted my changes and duplicated the block to open the URL in the browser

commands/local_server_start.go
```diff
@@ -274,11 +275,26 @@ var localServerStartCmd = &console.Command{
...
+			}
+		}
+
+		if config.Open {
+			if err := open.Run(applicationUrl); err != nil {
+				terminal.Eprintln("<error>Error while opening:", err, "</>")
+				terminal.Eprintfln("Please visit <href=%s>%s</> manually.", applicationUrl, applicationUrl)
+			} else {
+				terminal.Eprintfln("Opened: <href=%s>%s</>", applicationUrl, applicationUrl)
+			}
```
tucksaun:
```suggestion
			abstractOpenCmd(applicationUrl);
```

should be enough. or did I miss something?

tucksaun:
It is if you are the same package (ie. directory)

commands/local_server_start.go
Kocal: Indeed it works, fixed

fabpot (approved):

PR #669, opened 2025-11-11
Add support for `.symfony.cli.yaml` configuration file

Kocal:
As discussed internally some months ago, I suggested to add support for `.symfony.cli.yaml` file.

In 2025, such a name reflects better the purpose of the file, which is to configure the Symfony CLI tool for a given Symfony project. It's more user-friendly rather than `.symfony.local.yaml`, which IIRC is something _legacy_, and may be confusing.

The PR adds support for the following files:
- `.symfony.cli.dist.yaml`
- `.symfony.cli.yaml`
- `.symfony.cli.override.yaml`

### DIFF
--- a/local/project/config.go
+++ b/local/project/config.go
@@ -20,12 +20,14 @@
 package project
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
 	"github.com/symfony-cli/console"
+	"github.com/symfony-cli/terminal"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


local/project/config.go
```diff
@@ -78,16 +76,18 @@ func NewConfigFromDirectory(logger zerolog.Logger, homeDir, projectDir string) (
...
 	for _, suffix := range []string{".dist.yaml", ".yaml", ".override.yaml"} {
-		fileConfig, err := newConfigFromFile(filepath.Join(projectDir, ConfigFilePrefix+suffix))
-		if errors.Is(err, os.ErrNotExist) {
-			continue
-		} else if err != nil {
-			return nil, err
-		} else if fileConfig == nil {
-			continue
-		}
+		for _, configFilePrefix := range []string{".symfony.local", "symfony-cli", ".symfony-cli"} {
```
tucksaun: I think I would move this as the outerLoop. This way we can break as soon as we have a a match on one prefix and not let users mix everything together

tucksaun:
I don't think having both `symfony-cli` AND `.symfony-cli` make senses, it also adds confusion IMHO.
Using a dot prefixed file is my personal preference as this the convention most tools are using nowadays.
If the reason behind this is offering alternatives to users then we might instead take inspiration on Docker Compose and add support for configuring this via a flag and a environment variable.

Now regarding the out-of-the-box naming I'm not against getting rid of `.symfony.local` but I would like to have the whole Symfony Core Team(s) opinion as this will affect the ecosystem as a whole.

tucksaun:
And to be considered: there's the possibility for a `.symfony.insight.yaml` file to exists so we might want to consider `.symfony.cli.yaml`.

And this file is also configuring the behavior of the "Local" web server so some might prefer `.symfoy.local.yaml` 🤓

PR #694, opened 2026-01-21
Add support for hidden .php.ini configuration file

nacholibre:
### Problem
When using `frankenphp php-cli` (in my case in production) and there is `php.ini` file in the project root directory it forces franken to use that instead of the system default, causing unexpected behavior.

### Solution
My proposed solution is to make `symfony-cli` search for both php.ini and .php.ini (in that order). Users can rename their config to .php.ini so symfony-cli still finds it, but FrankenPHP ignores it.

### DIFF
--- a/local/php/executor.go
+++ b/local/php/executor.go
@@ -630,4 +630,3 @@ func (e *Executor) phpiniForDir() (string, string) {
 	}
 	return "", ""
 }
-
--- a/local/php/executor.go
+++ b/local/php/executor.go
@@ -275,9 +275,9 @@ func (e *Executor) Config(loadDotEnv bool) error {
 		// see https://php.net/manual/en/configuration.file.php
 		// if PHP_INI_SCAN_DIR exists, just append our new directory
 		// if not, add the default one (empty string) and then our new directory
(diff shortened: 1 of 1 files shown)


tucksaun:
Would this work? I mean: does PHP supports a `.php.ini` by default?

nacholibre:
> Would this work? I mean: does PHP supports a `.php.ini` by default?

What do you mean? AFAIK php doesn't load `php.ini` by default either. 

When using `symfony` cli, the `local/php/executor.go` is responsible for loading the `php.ini` config.

fabpot:
> > Would this work? I mean: does PHP supports a `.php.ini` by default?
> 
> What do you mean? AFAIK php doesn't load `php.ini` by default either.
> 
> When using `symfony` cli, the `local/php/executor.go` is responsible for loading the `php.ini` config.

`executor.go` does not load the ini file itself. We only append the project directory to PHP_INI_SCAN_DIR.
To answer Tug's question, yes as it actually loads all files ending with .ini.

So, if there are more than one .ini file, it will load all of them.

Issue #588, opened 2025-03-23
Command aliases - e.g. symfony migrate


weaverryan:
I'd love to see an aliases section in the config file where we can setup shortcuts, for example symfony migrate instead of symfony console doctrine:migrations:migrate.

We would show (per #582) that the config file is being loaded. We'd also sho the actual symfony command being run.

This seems like a low-hanging DX improvement and would be another reason to use the binary.

Also the symfony flex config configurator we've talked about can add these.

Thanks!

weaverryan:
Proposal updated above!

tucksaun:
I like this! But what do others (in particular @fabpot ) think about it?

weaverryan:
> There's also an in-between situation to address: should we first try to fuzzy match command names and display a suggestion if the command name is close to something we know?

Is there any downside? I cant think of one, so I'd say 👍

Kocal:
I think this suggestion will like to misunderstand. 

If in our documentations we start showing `symfony doctrine:migrations:migrate` instead of `symfony console doctrine:migrations:migrate`, then people will most likely expect `symfony` to always be “equivalent” to `bin/console`. 
If people were used to run `symfony console c:cl` or `bin/console c:cl` to clear the app cache, running `symfony c:cl` will instead execute `symfony cloud:clear-cache` which is totally different. 

How can we handle such a case?

weaverryan:
> If in our documentations we start showing symfony doctrine:migrations:migrate instead of symfony console doctrine:migrations:migrate, then people will most likely expect symfony to always be “equivalent” to bin/console.
If people were used to run symfony console c:cl or bin/console c:cl to clear the app cache, running symfony c:cl will instead execute symfony cloud:clear-cache which is totally different.

That's fair and we should probably print something like:

```bash
symfony c:cl
> Running "symfony cloud:clear-cache"
```

I think these edge cases are not worth keeping the current situation

```bash
artisan migrate
symfony console doctrine:migrations:migrate
```

PR #700, opened 2026-03-27
Feature #588 make unknown commands fallback to symfony console

ayyoub-afwallah:
Adds the requested feature in #588

 When a command is not recognized by the symfony CLI, it now automatically proxies to `symfony console <cmd>` if a `bin/console` is found in the project

```
  $ symfony debug:container
  Running "symfony console debug:container"
```

  - Unknown command + bin/console present → proxies to symfony console <cmd> and prints what it's running
  - Unknown command + no bin/console → falls through to the existing "unknown command" error with suggestions (no regression)
  - Known CLI commands (e.g. server:start, cloud:*) → unchanged
  - Abbreviated known commands (e.g. c:cl → cloud:clear-cache) → unchanged, abbreviation matching still applies

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/symfony-cli/console"
@@ -122,6 +123,14 @@ func main() {
 		Action: func(ctx *console.Context) error {
 			if ctx.Args().Len() == 0 {
 				return commands.WelcomeAction(ctx)
+			}
+			// Unknown command? fall back to "symfony console <cmd>" if a bin/console is found.
+			// e.g. "symfony debug:container" -> "symfony console debug:container"
+			cmdArgs := ctx.Args().Slice()
+			if executor, err := php.SymfonyConsoleExecutor(terminal.Logger, cmdArgs); err == nil {
+				terminal.Eprintfln("<info>Running \"symfony console %s\"</>", strings.Join(cmdArgs, " "))
+				executor.ExtraEnv = getCliExtraEnv()
+				return console.Exit("", executor.Execute(false))
 			}
 			return console.ShowAppHelpAction(ctx)
 		},


tucksaun (changes requested):
Thank you for starting to work on this!

Though I believe this implementation kind of duplicates the logic contained in the `run` method on `Application` and bypasses additional logic that might run via extension points.

Another issue I suspect is that it won't handle having flags before the command name.

So instead, I would rely on the default action (at lines 122-127) that is triggered when no command matched and when all the rest of the logic has already been handled

main.go
```diff
@@ -123,6 +124,14 @@ func main() {
 			if ctx.Args().Len() == 0 {
 				return commands.WelcomeAction(ctx)
 			}
+			// Unknown command? fall back to "symfony console <cmd>" if a bin/console is found.
+			// e.g. "symfony debug:container" -> "symfony console debug:container"
+			cmdArgs := ctx.Args().Slice()
+			if executor, err := php.SymfonyConsoleExecutor(terminal.Logger, cmdArgs); err == nil {
+				terminal.Printfln("<info>Running \"symfony console %s\"</>", strings.Join(cmdArgs, " "))
```
tucksaun, line 131: let's reuse the available helpers

```suggestion
				terminal.Printfln("<info>Running \"%s\"</>", executor.CommandLine())
```

tucksaun (approved):
Some nitpicking but otherwise seems ok to me

main.go
ayyoub-afwallah: good idea but using the executor would print the whole path to the console  `Running "php /root/project/bin/console debug:container"`.

maybe we can simply keep the current implementation?

ayyoub-afwallah:
@tucksaun I think this one is ready! otherwise, is there anything i could do to push it forward?

tucksaun (approved):

main.go
fabpot: Should probably be on stderr:

```suggestion
				terminal.Eprintfln("<info>Running \"symfony console %s\"</>", strings.Join(cmdArgs, " "))
```

main.go
```diff
@@ -123,6 +124,14 @@ func main() {
...
+				executor.ExtraEnv = getCliExtraEnv()
+				os.Exit(executor.Execute(false))
```
fabpot: Just a suggestion:

```suggestion
				return console.Exit("", executor.Execute(false))
```

main.go
ayyoub-afwallah: done

main.go
ayyoub-afwallah: done

fabpot:
Thank you @ayyoub-afwallah.

PR #708, opened 2026-05-08
Fix Windows file lock error when starting server in daemon mode

### LINKED ISSUE #349: Local server does not exit cleanly with CTRL+C

### LINKED ISSUE #631: All versions since 5.11.0 are broken on Windows
Using symfony cli on Windows 11 with php version 8.2.12.

This command worked fine on symfony cli version 5.11:

> symfony local:server:start --port=8000 --document-root=public_html -d

When I run it on version 5.12 it fails with this error message:

> remove C:\Users\ovgra\.symfony5\log\02dbcdfb26de9d3e05ec236c81e8227c736c1c30.log: The process cannot access the file because it is being used by another process.

If I omit the '-d' parameter the command does start the server, but when I stop the server with '^C' it throws the same error message one or many times.

Taminoful:
Fixes #631. Also fixes #349.

Credit to @Dwza for the investigation in [this comment](https://github.com/symfony-cli/symfony-cli/issues/631#issuecomment-2914275059), which captured the full stack trace and pinpointed `cleanupWebServerFiles` calling `pidfile.Remove()` as the source of the error. That stack trace was the blueprint for this fix.

## Root cause

On Windows, deleting an open file is not allowed. The server holds log files open via a log writer throughout its lifetime. Several startup and shutdown paths called `Remove()` on PID files, which also deletes the associated log file, causing the error:

```
remove ...\symfony-cli\log\<hash>.log: The process cannot access the file because it is being used by another process.
```

This affected two distinct scenarios:

**1. server:start -d (startup, introduced in 5.12, reported in #631)**

The parent opens the log file via `LogWriter()` before spawning the child via `reexec.Background()`. The child re-runs the full `Action` from scratch and calls `cleanupWebServerFiles()`, which tries to delete the same log file the parent still has open.

**2. server:stop and CTRL+C (shutdown, reported in #349)**

`WaitForExit()`, `Stop()`, `doAll()`, and `waitForWorkers()` all called `Remove()` during shutdown while runner goroutines still had log files open.

## Fix

**For the startup case:** guard `cleanupWebServerFiles()` so only the parent process runs it. The child skips it because the parent already performed cleanup before opening the log file.

```go
if !reexec.IsChild() {
    if err := cleanupWebServerFiles(projectDir, pidFile); err != nil {
        return err
    }
}
```

**For the shutdown case:** stop trying to delete log files during shutdown. `WaitForExit()`, `Stop()`, `doAll()`, and `waitForWorkers()` now only remove the PID file. Log files are opened with `O_TRUNC` on next server start so leaving them in place between runs is harmless.

## Testing

Built a Windows binary from the patched source and verified that:
- `symfony server:start -d` completes without error on Windows 10
- `symfony server:stop` completes without error
- CTRL+C shuts down cleanly without the file lock error
- `symfony server:start` works correctly again afterwards

Build yourself locally on WSL (or similar) with `GOOS=windows GOARCH=amd64 go build -o symfony.exe .`

### DIFF
--- a/commands/local_proxy_start.go
+++ b/commands/local_proxy_start.go
@@ -160,7 +160,7 @@ var localProxyStartCmd = &console.Command{
 			terminal.RemapOutput(lw, lw).SetDecorated(true)
 			reexec.NotifyForeground(reexec.UP)
 		} else {
-			defer pidFile.Remove()
+			defer os.Remove(pidFile.PidFile())
 		}
 
 		shutdownCh := make(chan bool, 1)
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -78,8 +78,13 @@ var localServerStartCmd = &console.Command{
 			ui.Warning("The local web server is already running")
 			return errors.WithStack(printWebServerStatus(projectDir))
 		}
-		if err := cleanupWebServerFiles(projectDir, pidFile); err != nil {
-			return err
+		// Only the parent cleans up stale files. The child skips this because
+		// the parent already ran cleanup and still holds the log file open;
+		// on Windows, deleting an open file is not allowed.
+		if !reexec.IsChild() {
+			if err := cleanupWebServerFiles(projectDir, pidFile); err != nil {
+				return err
+			}
 		}
 
 		homeDir := util.GetHomeDir()
--- a/commands/local_server_start.go
+++ b/commands/local_server_start.go
@@ -458,7 +458,7 @@ func waitForWorkers(projectDir string, pidFile *pid.PidFile) error {
 	if err := g.Wait(); err != nil {
 		return err
 	}
-	if err := pidFile.Remove(); err != nil {
+	if err := os.Remove(pidFile.PidFile()); err != nil && !os.IsNotExist(err) {
 		return err
 	}
 	return nil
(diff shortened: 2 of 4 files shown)


fabpot:
Added two missing similar problems.

fabpot:
Thank you @Taminoful.

tucksaun (commented):
its seems that those changes are breaking the PidFile encapsulation and uniformity in the API: the caller now need s to now that the pidFile store its state and logs as files ((eg. `os.Remove(pidFile.LogFile())` and `os.Remove(pidFile.PidFile())`) and this leaks implementation details leading to some edge case handling in the caller (`&& !os.IsNotExist(err)`) but sometimes it does not (eg. `RemoveWorkerLogs` does not)

I think we should stick with encapsulating the details or not but nothing something in between

fabpot:
> its seems that those changes are breaking the PidFile encapsulation and uniformity in the API: the caller now need s to now that the pidFile store its state and logs as files ((eg. `os.Remove(pidFile.LogFile())` and `os.Remove(pidFile.PidFile())`) and this leaks implementation details leading to some edge case handling in the caller (`&& !os.IsNotExist(err)`) but sometimes it does not (eg. `RemoveWorkerLogs` does not)
> 
> I think we should stick with encapsulating the details or not but nothing something in between

See #720

PR #711, opened 2026-06-02
fix: Allow stopping the server for removed directories

JoppeDC:
When developing with git worktrees, I spin up a server per worktree. After removing a worktree, the server process keeps running — but `symfony server:stop --dir <path>` could no longer kill it, because the directory it referenced was gone. The orphaned server was effectively unstoppable (short of finding and killing the PID by hand).

`getProjectDir` resolves the path via `filepath.EvalSymlinks`, which returns `fs.ErrNotExist` for a deleted directory. `server:stop` only needs the directory as a key to look up the server's PID files — it never reads from the directory itself — so failing here is unnecessarily strict.

### DIFF
--- a/commands/local_server_stop.go
+++ b/commands/local_server_stop.go
@@ -21,8 +21,11 @@ package commands
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
+	"path/filepath"
 
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @JoppeDC.

PR #724, opened 2026-07-25
Make error page commands easier to copy (minor DX)

smnandre:
| Before | After |
| - | - | 
| before | after |

The suggested command text contained the `$`, so selecting the line 
to copy it picked up the `$` too.

This PR now write the `$` via a CSS `::before` pseudo-element: still visible, 
but out of the DOM text (and `user-select: none` for older browsers).

Same change applied to the other error page (hostname not linked).

### DIFF
--- a/local/html/html.go
+++ b/local/html/html.go
@@ -186,6 +186,11 @@ a, a.link, a:visited, a:hover {
     width: 7px;
     height: 7px;
 }
+.terminal .command:before {
+	content: "$ ";
+	-webkit-user-select: none;
+	user-select: none;
+}
 .button {
 	font-family: monospace;
 	background-color: #42a7ff;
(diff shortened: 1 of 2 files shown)


local/html/html.go
```diff
@@ -186,6 +186,11 @@ a, a.link, a:visited, a:hover {
     width: 7px;
     height: 7px;
 }
+.terminal .command:before {
+	content: "$ ";
+	-webkit-user-select: none;
```
Kocal, line 191:
```suggestion
```

Support without vendor prefix has been added in Chrome 54

smnandre: Indeed, but not in Safari https://caniuse.com/user-select-none  :|

smnandre:
(not sure why the CI fail, "internal error" and I cannot relaunch it)

local/html/html.go
Kocal: Ooooh, no way! 🤯

fabpot:
Thank you @smnandre.

PR #729, opened 2026-08-02
Stream FastCGI responses that have no Content-Length

### LINKED ISSUE #728: Streamed responses still buffer unless the content type is text/event-stream
## Summary

A `StreamedResponse` served by `symfony serve` reaches the client in 4KB blocks
instead of as it is written, unless its content type is `text/event-stream`.

This is the part of #330 that #371 did not cover. SSE works, which is why the
issue has looked resolved since 2023 — but NDJSON, `text/plain`, and any other
streamed content type still buffer.

## Reproduction

```php
#[Route('/stream-probe')]
public function __invoke(Request $request): StreamedResponse
{
    $response = new StreamedResponse(static function (): void {
        $started = microtime(true);
        for ($i = 1; $i <= 4; ++$i) {
            printf("frame %d at +%.2fs\n", $i, microtime(true) - $started);
            flush();
            sleep(1);
        }
    });
    // Swap for text/event-stream and the very same code streams correctly.
    $response->headers->set('Content-Type', 'text/plain; charset=utf-8');

    return $response;
}
```

```bash
curl -sN https://127.0.0.1:8000/stream-probe
```

Read the response raw — piping through `head`/`grep`/`awk` reads in blocks and
makes a correctly streaming response look buffered.

**Expected:** four lines, one per second.
**Actual:** all four lines at once, after four seconds.

Same code streams correctly with `php -S`, and behind nginx + PHP-FPM.

## Evidence that the buffer is size-bound and lives in the proxy

Padding each frame to ~2KB makes the frames arrive in *pairs*, at exactly the
4KB boundary — arrival is triggered by bytes written, with no time component:

```
+1.24s  frame 1 ...   <- frames 1 and 2 together, once ~4KB accumulated
+1.24s  frame 2 ...
+3.24s  frame 3 ...   <- frames 3 and 4 together, at the next 4KB
+3.24s  frame 4 ...
```

The response headers are delayed with the body, and the unpadded response
arrives carrying `Content-Length: 192` — a header a `StreamedResponse` never
sets, which means the body was fully accumulated and then measured by Go.

## Cause

`local/fcgi_client/fcgiclient.go`:

```go
resp.ContentLength, _ = strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
```

With no `Content-Length` header, `ParseInt("")` fails, the error is discarded,
and `resp.ContentLength` is left at **0** rather than **-1** — so an unknown
length is reported as an empty body of known length.

`httputil.ReverseProxy` then decides flushing from that field:

```go
func (p *ReverseProxy) flushInterval(res *http.Response) time.Duration {
	resCT := res.Header.Get("Content-Type")
	if baseCT, _, _ := mime.ParseMediaType(resCT); baseCT == "text/event-stream" {
		return -1
	}
	// We might have the case of streaming for which Content-Length might be unset.
	if res.ContentLength == -1 {
		return -1
	}
	return p.FlushInterval
}
```

`text/event-stream` hits the first branch and streams. Everything else should
hit the second one, but cannot, because `ContentLength` is 0. It falls through
to `p.FlushInterval`, which `local/php/php_server.go` never sets — so 0, meaning
never flush.

That second branch exists precisely to cover the "all cases" that #330 discussed
but #371 left open; #371 touched `cgi.go`, `context.go`, `php_server.go`,
`toolbar.go` and `xsendfile.go`, but not the FastCGI client where the field is
populated.

Setting `ContentLength = -1` when the header is absent makes both PHP-FPM and
PHP-CGI stream any content type. PR follows.

## Environment

- Symfony CLI 5.17.1, and reproduced against `main` (c15ea4a)
- PHP 8.5.4, FPM (also applies to the CGI path — both share `cgiTransport`)
- macOS 15 / arm64

TNAJanssen:
Fixes #728.

A response with no `Content-Length` header has an *unknown* length, which
net/http spells as `-1`. `ParseInt("")` fails and its error is discarded, so
`resp.ContentLength` was left at `0` — "an empty body of known length".

`httputil.ReverseProxy` decides flushing from exactly that field:

```go
if baseCT == "text/event-stream" { return -1 }  // flush every write
if res.ContentLength == -1       { return -1 }  // flush every write
return p.FlushInterval                           // never set here, so 0 = never flush
```

So a streamed response only bypassed buffering when it was SSE. Everything else
took the third line and reached the client in 4KB blocks, or at the end of the
request — the remainder of #330 after #371, which never touched this file.

Setting the field to `-1` when the header is absent lets the second branch do
what it is there for, for both the FPM and CGI paths (they share `cgiTransport`).

### Verification

With PHP 8.5.4 FPM and a `StreamedResponse` writing one line per second:

| Content type | before | after |
|---|---|---|
| `text/plain` | 4 lines at once, after 4s | 1 line per second |
| `text/event-stream` | 1 line per second | 1 line per second |

Before the change the unpadded `text/plain` response also arrived carrying a
`Content-Length` header that PHP never sent, Go having measured the body it had
accumulated. Padding each frame to ~2KB made frames arrive in pairs on the 4KB
boundary, confirming the buffer was size-bound rather than time-bound.

Non-streamed responses are unaffected in practice: they already flow through the
same proxy, and the Symfony profiler/toolbar injection (`ModifyResponse`) still
rewrites HTML responses correctly — verified against a 296KB profiler page.

`gofmt`, `go vet ./...` and `go test ./...` are clean.

### DIFF
--- a/local/fcgi_client/fcgiclient.go
+++ b/local/fcgi_client/fcgiclient.go
@@ -366,7 +366,15 @@ func (f *FCGIClient) Request(p map[string]string, req io.Reader) (resp *http.Res
 	resp.Header = http.Header(mimeHeader)
 	// TODO: fixTransferEncoding?
 	resp.TransferEncoding = resp.Header["Transfer-Encoding"]
-	resp.ContentLength, _ = strconv.ParseInt(resp.Header.Get("Content-Length"), 10, 64)
+	// A response without a Content-Length header has an unknown length, which
+	// net/http spells as -1. Leaving it at 0 makes httputil.ReverseProxy take a
+	// streamed response for a complete one: its flushInterval() only bypasses
+	// buffering for text/event-stream or for an unknown length, so any other
+	// streamed content type reaches the client in 4KB blocks instead.
+	resp.ContentLength = -1
+	if contentLength := resp.Header.Get("Content-Length"); contentLength != "" {
+		resp.ContentLength, _ = strconv.ParseInt(contentLength, 10, 64)
+	}
 
 	// status
 	err = errors.WithStack(extractStatus(resp))
--- a/local/fcgi_client/fcgiclient_test.go
+++ b/local/fcgi_client/fcgiclient_test.go
@@ -1,0 +1,75 @@
+package fcgiclient
+
+import (
+	"bytes"
+	"encoding/binary"
+	"io"
+	"testing"
+)
+
+func TestRequestContentLength(t *testing.T) {
+	tests := []struct {
+		name    string
+		headers string
+		want    int64
+	}{
+		{
+			name:    "unknown length",
+			headers: "Content-Type: text/plain\r\n",
+			want:    -1,
+		},
+		{
+			name:    "known length",
+			headers: "Content-Type: text/plain\r\nContent-Length: 5\r\n",
+			want:    5,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			client := newResponseClient(t, test.headers+"\r\nhello")
+
+			response, err := client.Request(nil, nil)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if response.ContentLength != test.want {
+				t.Fatalf("got content length %d, want %d", response.ContentLength, test.want)
+			}
+		})
+	}
+}
+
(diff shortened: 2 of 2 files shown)


fabpot:
Thank you @TNAJanssen.

Issue #522, opened 2024-10-08
critical CVE-2024-41110 rated 9.9


david-maumenee:
In our custom PHP Docker image built the 26th of September, Trivy scan detected the critical [CVE-2024-41110](https://avd.aquasec.com/nvd/cve-2024-41110) rated 9.9.

```
usr/local/bin/symfony (gobinary)
================================
Total: 2 (HIGH: 1, CRITICAL: 1)

┌──────────────────────────┬────────────────┬──────────┬────────┬──────────────────────┬─────────────────────────────────┬───────────────────────────────────────────────────────────┐
│         Library          │ Vulnerability  │ Severity │ Status │  Installed Version   │          Fixed Version          │                           Title                           │
├──────────────────────────┼────────────────┼──────────┼────────┼──────────────────────┼─────────────────────────────────┼───────────────────────────────────────────────────────────┤
│ github.com/docker/docker │ CVE-2024-41110 │ CRITICAL │ fixed  │ v27.0.3+incompatible │ 23.0.15, 26.1.5, 27.1.1, 25.0.6 │ moby: Authz zero length regression                        │
│                          │                │          │        │                      │                                 │ https://avd.aquasec.com/nvd/cve-2024-41110                │
├──────────────────────────┼────────────────┼──────────┤        ├──────────────────────┼─────────────────────────────────┼───────────────────────────────────────────────────────────┤
```

docker-ce `v27.1.1` containes patches to fix the vulnerability. @fabpot upgraded this library to a fixed version in this [commit](https://github.com/symfony-cli/symfony-cli/commit/367ab89bd03a1404fe92c59194d33e3e6f1cd5f8) the 27th of August. 

But the last [tag](https://github.com/symfony-cli/symfony-cli/releases/tag/v5.10.2) was release in July.

Could you release a new symfony-cli's version in order to fix this critical CVE ?

fabpot:
Release done.

tucksaun:
For the record, the mentioned CVE applies to the server part of the Docker Engine which is not used by the Symfony CLI.

PR #732, opened 2026-08-07
Update golang.org/x/text to v0.39.0 — last dependency with a released security fix

mathieu-ducrot:
## Context

  While scanning our PHP development container images with `osv-scanner` (v2.4.0), we noticed the latest released binary (**v5.17.1**, built 2026-04-07 with Go 1.26.0) reports **56 known vulnerabilities** in its embedded
  dependencies — including GO-2026-5023 (CVSS 10.0) in `golang.org/x/crypto` v0.49.0.

  Digging further, we found that `main` has already moved past most of these since the v5.17.1 tag (`x/crypto` v0.53.0, `x/net` v0.56.0, `otel` v1.44.0, `go-pkcs12` v0.7.3…). The only remaining module carrying an advisory with a
  released fix was `golang.org/x/text` v0.38.0.

  ## This PR

  Bumps `golang.org/x/text` v0.38.0 → v0.39.0, fixing GO-2026-5970.

  - `go build .` ✅
  - `go test ./...` ✅ (all packages pass)
  - `go vet ./...` ✅

  ## Measured impact

  We rebuilt the binary from this branch with Go 1.26.5 and scanned it:

  | | v5.17.1 release binary | Build from this branch (Go 1.26.5) |
  |---|---|---|
  | Advisories reported | 56 (7 Critical, incl. one CVSS 10.0) | **6 (0 Critical)** |
  | With a released fix | 50 | **0** |

  The 6 remaining advisories have no released fix anywhere: 5 are `github.com/docker/docker` daemon-side issues (not addressable through the Go client module) and 1 is GO-2026-5932 (`x/crypto`, no upstream fix yet).

  ## Friendly ask

  With this merged, `main` is as clean as it can currently get. Would you consider **tagging a release** built with Go ≥ 1.26.5? That alone would take the binary shipped to users (and embedded in countless container images, where
  every security scan flags it) from 56 advisories down to 6 non-actionable ones. You have kindly done such releases before (#522, #485, #468).

  Thanks for maintaining the CLI!


tucksaun (approved):

tucksaun:
as far as I can tell, GO-2026-5932 is about `golang.org/x/crypto/openpgp` which I don't think we are using at the moment.

regarding this precise bump looks good to me 👍

tucksaun (approved):

tucksaun:
could you also check with `govulncheck` instead?

mathieu-ducrot:
@tucksaun  I tested this with `govulncheck` as requested.

  **Setup:** govulncheck v1.6.0, Go vuln DB from 2026-07-27. I ran `govulncheck ./...`
  on `main` and on this branch to compare.

  **Result:**
  
  | | main (x/text v0.38.0) | this branch (v0.39.0) |
  |---|---|---|
  | Called vulnerabilities | 21, from 2 modules | 20, from 1 module |

  The only difference is `GO-2026-5970` (infinite loop on invalid input in
  golang.org/x/text, fixed in v0.39.0). It's gone on this branch, and nothing new
  was introduced.

  govulncheck also confirms the vulnerability was actually reachable in the code,
  not just present in go.mod:

      commands/local_server_ca_install.go:76:22: commands.init calls cert.CA.MakeCert,
        which eventually calls norm.Form.Bytes
      local/php/toolbar.go:149:198: php.Server.tweakToolbar calls cases.Caser.String,
        which eventually calls norm.Form.Properties

  **Two unrelated things the scan surfaced** (out of scope for this PR, just flagging):

  1. **17 stdlib vulnerabilities** (crypto/tls, crypto/x509, archive/tar, os, net/url…),
     fixed across go1.26.1 → go1.26.5. My local toolchain was go1.26.0, so this may
     just be my machine — but since CI uses `go-version-file: 'go.mod'` (`go 1.26.0`),
     it might be worth checking which patch release setup-go actually resolves to for
     release builds. These are invisible to OSV/Dependabot since they aren't dependencies.

  2. **3 vulnerabilities in `github.com/docker/docker@v28.5.2+incompatible`** with no
     fixed version available. They look like false positives for us: they affect the
     Moby daemon, while `envs/docker.go` only uses the client.

  Happy to paste the full reports if useful.

fabpot (approved):

fabpot:
Thank you @mathieu-ducrot.

